PR #172, opened 2020-10-15
Implment ISIS Specific Endpoints for Python ICAT Backend

### LINKED ISSUE #146: implement ISIS-specific endpoints for python-icat
- [ ]  get_instrument_facilitycycles_with_filters
- [ ]  count_instrument_facilitycycles_with_filters
- [ ] get_instrument_facilitycycle_investigations_with_filters
- [ ]  count_instrument_facilitycycles_investigations_with_filters

MRichards99:
This PR closes #146.

This adds functionality to the 4 ISIS endpoints in this API (2 return lists of results, 2 are count queries). I've tested these endpoints on the ISIS dev ICAT and they match the results from the TopCat queries. If you test against that ICAT instance, make sure you create a new session ID using this API otherwise you'll just get 403s. This is because you have to log in the `client` object currently, but this will be fixed with #156 (I've just tested it works properly on that branch, as I only tested that on my dev instance beforehand!). Bit of a pain, but at least there's a proposed solution in place :)

When using WHERE filters in this API, you can now specify attributes (of the attributes of the query, included entities not supported on RHS) on the RHS of the expression. The code to make this possible can be found below:
https://github.com/ral-facilities/datagateway-api/blob/16520486550a9a4051c1297c99106f22c7a159ab/common/icat/filters.py#L78
An example of this feature being used can be seen here:
https://github.com/ral-facilities/datagateway-api/blob/16520486550a9a4051c1297c99106f22c7a159ab/common/icat/helpers.py#L584

`ICATQuery` now has an attribute to specify whether the query is for an ISIS endpoint or not. ISIS endpoints require DISTINCT aggregates on the queries so duplicate data isn't returned, however this is different to a distinct field filter that is seen in the API, so this attribute stops code regarding distinct field filters from executing when only an aggregate is specified on an ISIS endpoint.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,6 @@
+**Description**:
+Add a description here
+
+**Acceptance criteria**:
+- [ ] A checklist
+- [ ] of criteria to accept this story as done
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,15 @@
+This PR will close #{issue number}
+
+## Description
+Enter a description of the changes here
+
+## Testing instructions
+Add a set up instructions describing how the reviewer should test the code
+
+- [ ] Review code
+- [ ] Check Travis build
+- [ ] Review changes to test coverage
+- [ ] {more steps here}
+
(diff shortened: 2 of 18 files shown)


MRichards99:
Merging this as there's only the timezone stuff to look at on #171

PR #175, opened 2020-10-16
Improve Logging Throughout Python ICAT Backend

### LINKED ISSUE #164: Improve Logging throughout Python ICAT Backend
Over the past couple of issues I've been working on, I've added various `log.info()` (to show what the API is doing) and `log.debug()` (to show what values are inside the API at the time) where relevant. I'd like to go back to some of the older code I've written for this backend and add some logging. I'll do this once most of my PRs have been merged into `master`, just to avoid merge conflicts and the like.

MRichards99:
This PR will close #164.

This is a very simple PR to improve the logging throughout the Python ICAT backend. I've been better at this more recently, but some of the earlier development didn't have much logging in it. It's not overly detailed to prevent the log files from being bloated, but it gives information about what the user is doing within the API.

### DIFF
--- a/common/icat/helpers.py
+++ b/common/icat/helpers.py
@@ -481,7 +481,7 @@ def get_facility_cycles_for_instrument(
     """
     # TODO - Add logging
 
-    query_aggregate = "COUNT" if count_query else None
+    query_aggregate = "COUNT:DISTINCT" if count_query else "DISTINCT"
     query = ICATQuery(
         client, "FacilityCycle", aggregate=query_aggregate, isis_endpoint=True
     )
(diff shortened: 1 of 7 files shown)


common/icat/helpers.py
```diff
@@ -84,7 +84,7 @@ def logout_icat_client(client):
     :param client: ICAT client containing an authenticated user
     :type client: :class:`icat.client.Client`
     """
-
+    log.info("Logging out of the Python ICAT client")
```
stuartpullinger, line 87: Why log here and not in backend.py like the other similar methods eg. login, refresh etc?

stuartpullinger (changes requested):

common/icat/helpers.py
MRichards99: Good spot of the inconsistency. I've only changed this instance because other instances of logging in `helpers.py` rather than `backend.py` are because the functions are used elsewhere (`get_entity_by_id()` is used when updating data for example). For this reason, only session related endpoints have logging in `backend.py`, so non-session endpoints are consistent in the fact their logging is in their respective helper functions.

stuartpullinger (approved):

PR #177, opened 2020-10-19
Add Postman Collection to Repo

MRichards99:
I've added the Postman Collection I've been building up over the last few months while developing the new backend. It contains 320 requests (I think every single endpoint plus a couple of duplicated ones that I use often so they're in a dedicated folder). More details can be seen in the commit message of cf83b3f2bc88e4b26064bfe552b5446474412962 but the JSON file should be ready to import into your own Postman instance and use. There are different folders to categorise the different types of endpoints, and a folder which contains the requests I use quite a lot.

### DIFF
--- a/postman_collection_icat.json
+++ b/postman_collection_icat.json
@@ -1,6 +1,6 @@
 {
 	"info": {
-		"_postman_id": "124fd4f3-2ce0-4860-ab79-fae94eaeec29",
+		"_postman_id": "c54f3f84-4e41-44cb-93a1-434f3d91d65e",
 		"name": "DataGateway API",
 		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
 	},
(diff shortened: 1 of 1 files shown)


VKTB:
Executing the `Login - ICAT (simple)` request or any other request was erroring initially. This was because the value of the collection's `datagateway-api` variable was set to `http://127.0.0.1` while the request's URL is `http://{{datagateway-api}}/sessions`. This means that when it got executed it was sending the request to `http://http://127.0.0.1:15000/sessions` - notice the double `http://` at the beginning. Also the port number is set to `15000` but the API ran on `5000` on my machine, so it might be worth adding something to the `README` about this. 

Once I re-configured the `datagateway-api` variable, I was able to execute the other requests but it's worth mentioning that the collection's `sessionID` variable did not automatically pick up the `sessionID` returned by the `Login - ICAT (simple)` request so I had to manually set it. There were some requests that returned `500 ` status codes but this is nothing to do with the collections.

MRichards99:
@VKTB I've made the changes requested, so please could you re-review? The issues you found with the variables was because I used environment variables, not collection variables. I've changed the setting of the `sessionID` to `127.0.0.1:5000` so there shouldn't be a duplicated http and the default port is set. The port was 15000 because I do port forwarding from my VM to my Windows laptop and adding a 1 on the port seems to be the default ICAT behaviour, but the Postman collection should assume the default port is used.

VKTB (approved):
Works nicely now

PR #178, opened 2020-10-21
Update Swagger/OpenAPI Docs for New Backend

### LINKED ISSUE #166: Fix Data Parsing within Swagger Docs
When I use the Swagger docs to check out details of the API, I find that the filters aren't passed correctly into the requests (see screenshot below). For example, a WHERE filter is parsed into the request URL as `where=[object Object]` which obviously isn't correct and returns a 400.

This should be investigated and fixed, as I'm not exactly sure what the problem is. 

image

MRichards99:
This PR will close #166. 

These changes look at the Swagger docs and the interface used to display the API in a friendly manner. There is one change per commit, so the changes on this branch should be pretty easy to follow but to summarise, there are the following types of changes:
- Corrections for the endpoints which have `id_` as a parameter
- Add functionality in DB backend for `ne` operation on a WHERE filter (this is to match the functionality seen in the ICAT backend)
- Add examples of all possible WHERE filter operations
- Adapt example values for filters to the ICAT backend (essentially this is just turning attributes to camelCase, rather than SNAKE_CASE)
- Committing the new YAML file as a result of these changes

### DIFF
--- a/common/database/filters.py
+++ b/common/database/filters.py
@@ -71,6 +71,8 @@ class DatabaseWhereFilter(WhereFilter):
 
         if self.operation == "eq":
             query.base_query = query.base_query.filter(field == self.value)
+        elif self.operation == "ne":
+            query.base_query = query.base_query.filter(field != self.value)
         elif self.operation == "like":
             query.base_query = query.base_query.filter(field.like(f"%{self.value}%"))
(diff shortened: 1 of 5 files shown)


stuartpullinger (approved):
Looks good.

PR #192, opened 2020-12-09
Add ICAT Backend Documentation

### LINKED ISSUE #190: Add documentation for ICAT Backend
A lot of `README.md` gives details regarding the DB backend, but I haven't really documented much about the ICAT backend. The PR linked to this issue should add documentation regarding the ICAT backend and the differences between the two. It's probably also a good idea for me to go through the file and clean up everything in there, ensure everything's up to date etc.

MRichards99:
This PR will close #190 

## Description
This PR makes quite a lot of change to `README.md` for the changes I've made to the API over the past few months. I've also cleaned up what was in the file beforehand.

There's a commit related to a suggested change which should be made in. I made the change in this branch for speed and simplicity, as explained: https://github.com/ral-facilities/datagateway-api/pull/185#discussion_r539409854

I've also renamed `query_filter.py` to `query_filter_factory.py` to make prevent any confusion. 

## Testing instructions

- [ ] Review documentation
- [ ] Review filename change

## Agile board tracking
connect to #190


README.md
```diff
@@ -249,53 +277,142 @@ pre-commit install
 ```
 
 
-## Running DataGateway API
+
+
+# Running DataGateway API
 Depending on the backend you want to use (either `db` or `python_icat`) the connection
```
patrick-austin, line 283: Could have "see more details [here]" style link to the section on the backend here?

README.md
```diff
@@ -249,53 +277,142 @@ pre-commit install
...
 
-Ideally, the API would be run with:
-`poetry run python -m datagateway_api.src.main`
-However it can be run with the flask run command as shown below:
+By default, the API will run on `http://localhost:5000` and all requests are made here
+e.g. `http://localhost:5000/sessions`.
+
 
+## API Startup
+Ideally, the API would be run using the following command:
```
patrick-austin, line 293: Is there a reason that it should be "ideally" done with `poetry` and not `flask`? Or is it just the point below about flask not taking the `config.json` options into account?

README.md
```diff
@@ -347,79 +465,204 @@ This is illustrated below.
...
+
+Each backend module contains the following files which offer similar functionality,
+implemented in their own ways:
+- `backend.py` - Implemented version of `datagateway_api.common.backend`
+- `filters.py` - Inherited versions of each filter defined in
+  `datagateway_api.common.filters`
+- `helpers.py` - Helper functions that are called in `backend.py`
+
+
+### Creating a Backend
```
patrick-austin, line 614: So I'm not 100% clear on this part, the documentation seems to imply that the backend will be created for me when running `main.py`, but I could only get it to load the swagger interface when I gave it the pre-prod url (`"ICAT_URL": "https://scigateway-preprod.esc.rl.ac.uk:8181"`). I suppose that I've misunderstood and that the API is created for me, but I would need to have actually created a database (or python backend) myself?

patrick-austin (changes requested):
On the whole I thought it was really well written, and I liked the use of external links and example commands.

There were one or two things I couldn't get to work, I don't know whether the root cause is elsewhere or if it's say a missing command from the documentation, but I thought it best to mention it at least. If it definitely should work based on the commands given and it's just a mistake on my end then that's fair enough.

README.md
```diff
@@ -120,13 +145,16 @@ intricacies of this command:
 poetry add [PACKAGE-NAME]
 ```
 
-### Nox (Automated Testing & Other Code Changes)
```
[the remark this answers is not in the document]
MRichards99: It's difficult to know for sure what issue you might be facing, but this is an interesting issue, I wonder if the bit about adding `--without-hashes` to the `poetry export` commands might work: https://github.com/python-poetry/poetry/issues/3472#issuecomment-744356551 

What do you mean by pinning versions in the functions in `noxfile.py`?

README.md
```diff
@@ -249,53 +277,142 @@ pre-commit install
...
+representation of the API will be shown using
+[Swagger UI](https://swagger.io/tools/swagger-ui/). This uses an OpenAPI specfication to
+visualise and allow users to easily interact with the API without building their own
+requests. It's great for gaining an understanding in what endpoints are available and
+what inputs the requests can receive, all from an interactive interface.
+
+This specification is built with the Database Backend in mind (attribute names on
+example outputs are capitalised for example), however the Swagger interface can also be
+used with the Python ICAT Backend. More details on how the API's OpenAPI specification
+is built can be found [here](TODO INSERT HYPERLINK).
```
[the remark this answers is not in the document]
MRichards99: Done, should point to another point within the README.

README.md
```diff
@@ -249,53 +277,142 @@ pre-commit install
...
+below:
 
 **Warning: the host, port and debug config options will not be respected when the API is
 run this way**
 
-To use `flask run`, the enviroment variable `FLASK_APP` should be set to `src/main.py`.
-Once this is set the API can be run with `flask run` while inside the root directory of
-the project. The `flask run` command gets installed with flask.
+To use `flask run`, the enviroment variable `FLASK_APP` should be set to
+`datagateway_api/src/main.py`. Once this is set, the API can be run with `flask run`
```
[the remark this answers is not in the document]
MRichards99: Good spot - I've corrected those paths.

README.md
MRichards99: Added an link to a section lower down in the README.

README.md
MRichards99: I wasn't entirely sure either (this section predates my time with the API), but I found out that `flask run` should only be run for development purposes. I've reworded this section and added a link to specifically where the Flask docs advise against production use of the command. Obviously ignoring `config.json` is also a reason too.

README.md
MRichards99: The concept of a backend is only a developer concept and isn't something that a user needs to know about. You should've been able to do something like the following though?

```bash
curl --request POST 'http://127.0.0.1:5000/sessions' --header 'Content-Type: application/json' --data '{
"username": "root",
"password": "pw",
"mechanism": "simple"
}'
```

README.md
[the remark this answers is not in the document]
patrick-austin: Adding `--without-hashes` [here](https://github.com/ral-facilities/datagateway-api/blob/feature/add-icat-backend-documentation-%23190/noxfile.py#L17) works like you said (I also noticed it's already present in [another command further down](https://github.com/ral-facilities/datagateway-api/blob/feature/add-icat-backend-documentation-%23190/noxfile.py#L77)).

I had thought another way round it would be to include the hashes in the `pip install` command which as far as I can tell is generated [here](https://github.com/ral-facilities/datagateway-api/blob/feature/add-icat-backend-documentation-%23190/noxfile.py#L49). For example replacing `"flake8"` with `"flake8==3.7.9 "`, which is what the 4th bullet point in the comment you link suggests. But since the `--without-hashes` method seems to work (and seems simpler) I guess it'd be better to go with that?

README.md
patrick-austin: Yeah I was able to submit requests through the swagger interface (through browser) or with `curl` after setting `"ICAT_URL": "https://scigateway-preprod.esc.rl.ac.uk:8181"`. My confusion was with whether I'd missed a step and should be say setting something up to run on `localhost:8181` while following the documentation and then provide that in the config file or whether someone following the documentation would already have an ICAT_URL instance set up for them and they only need to modify the config file. I guess from your response it's the latter?

MRichards99: I would imagine a 'typical' user would probably have an ICAT_URL already so they only need to modify the config file. As a developer, I have an ICAT instance setup locally which runs on `localhost:8181` (I used some tutorials on the [icat.manual](https://github.com/icatproject/icat.manual/tree/master/tutorials) repo) but any form of prod/pre-prod would use a well-established ICAT URL. I've added a link to these tutorials in the README, as per commit 73bb72d186a321b81bfba4e02484aed37d81e604

README.md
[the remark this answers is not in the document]
MRichards99: Thanks for testing that, I've committed that change, as per commit fb3900b25334aa141c36b33efb17bc71e137c043. Yeah that would be another way around the issue but I guess that means managing versions within the noxfile and within Poetry - not the end of the world since they're dev dependencies, but still nice to avoid. Hopefully this is now resolved.

README.md
patrick-austin: That makes sense, and I think the extra explanation helps clarify it (though I do appreciate that typical users and experienced devs probably wouldn't need to be explicitly told this).

patrick-austin (approved):
I'm happy with with the small changes to areas I didn't understand or couldn't get to run the first time round. While the additions might not be necessary for someone who knows what they're doing, as someone who's looking at this for the first time they helped me understand.

PR #194, opened 2021-01-07
Add NOT Like Operator to WHERE Filter

### LINKED ISSUE #193: Add NOT Like WHERE Filter
**Description**:
As per https://github.com/ral-facilities/datagateway/issues/484, a NOT like WHERE filter should be added to the API. I've quickly checked `python-icat` and I know this is feasible for that backend. I've tried to keep the filters (and general functionality) aligned between both backends so it would be good to add this functionality to the database backend too.

**Acceptance criteria**:
- [ ] On the Python ICAT Backend, a user should be able to add a NOT like where filter to any request that accepts filters, with the result returning a correct result
- [ ] The same functionality should also appear on the DB backend if it can be done by sqlalchemy
- [ ] An example usage should be added to the swagger docs
- [ ] Tests should be added in `test_where_filter.py` to test this version of the where filter

MRichards99:
This PR will close #193

## Description
As per the issue, there is a request in DataGateway for users to filter by data not containing a particular value. I've added this functionality on both backends by adding a new operator for the where filter. There are two extra tests that test this particular operator of the where filter too.

## Testing instructions
`http://{{datagateway-api}}/investigations?where={"TITLE": {"nlike":"dog"}}` should return a list of investigations where "dog" isn't in the title.

- [x] Review code
- [ ] Check the added tests pass

## Agile board tracking
connect to #193

### DIFF
--- a/datagateway_api/common/database/filters.py
+++ b/datagateway_api/common/database/filters.py
@@ -77,6 +77,8 @@ class DatabaseWhereFilter(WhereFilter):
             query.base_query = query.base_query.filter(field != self.value)
         elif self.operation == "like":
             query.base_query = query.base_query.filter(field.like(f"%{self.value}%"))
+        elif self.operation == "nlike":
+            query.base_query = query.base_query.filter(field.notlike(f"%{self.value}%"))
         elif self.operation == "lt":
             query.base_query = query.base_query.filter(field < self.value)
         elif self.operation == "lte":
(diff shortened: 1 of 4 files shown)


VKTB (approved):
Tested it with an ICAT backend and the results returned were expected and the `test_query_filter_factory.py` test passed successfully.

patrick-austin (approved):
Same as Viktor, works well as far as I can see.

PR #196, opened 2021-01-18
Implement GitHub Actions Workflow

### LINKED ISSUE #163: Implement CI into this repo
Like many of the other repos in SEG, it would be good to get some form of continuous integration into this codebase. Within the team, Travis is typically used for this, which requires the repo to be public, which currently isn't the case. I believe there's nothing stopping the repo becoming public, just when I choose to start implementing CI onto this repo. 

Since I've never used Travis, I'm a little unsure how Travis will get data from an ICAT server and if the unit tests will controlling the data (i.e. injecting and deleting test data) will be an issue on whichever ICAT server is chosen, but I'm sure someone else will understand how this all works.

MRichards99:
This PR will close #163

## Description
This PR implements a workflow for GitHub Actions to run when PRs are opened and when branches are merged into master. I've also added the `workflow_dispatch` keyword so you can run the workflow manually (through the website) however the workflow file must be in master for this to work.

In a summary, the workflow builds the API, builds a local instance of ICAT, injects sample data into `icatdb`, then runs the Nox sessions inside `noxfile.py`. The workflow is separated into separate jobs to easily diagnose which part of the workflow is causing issues. Tests are run on 3.6, 3.7, 3.8 and 3.9 - the API has dependencies which require at least Python 3.6, so Python 3.5 backwards cannot be tested.

I've updated `config.json.example` to some of the config options required for the Actions runner. These example values are still adaptable and similar for local/other usage so I don't think it causes any issues. I've updated the expected values on the config unit tests so they continue to pass. Any passwords that have been added (in the workflow file and the icatdb passwords) are already available publicly (in the icat-ansible repo) so I don't see any cause for security concern with those (they're just the standard ICAT formatted passwords). 

Some of the nox sessions were specified to run on Python 3.6 (they only needed to run on a single version, so I specified that because that's what prod uses), but I've removed this so they can be run on whatever Python version the user has. The sessions in question are for linting so don't actually execute API code.

The entire workflow takes just under 10 minutes to run, where jobs are parallelised.

## Testing instructions

- [ ] Review changes
- [ ] Check Actions build

## Agile board tracking
connect to #163


VKTB (approved):
Looks good. If dependencies ever grow and installations become slow, then the [Cache](https://github.com/actions/cache) GitHub action might help with caching dependencies between workflow runs.

VKTB:
@MRichards99 Just noticed but could you please update the [PULL_REQUEST_TEMPLATE.md](https://github.com/ral-facilities/datagateway-api/blob/feature/code-coverage-%23167/.github/PULL_REQUEST_TEMPLATE.md) file to say _Check Actions build_ instead of _Check Travis build_.

MRichards99:
@VKTB good spot - I've just pushed a commit updating the PR template

VKTB (approved):

PR #199, opened 2021-01-21
Merge Remove SQL Dependency PR into Master

MRichards99:
The original PR (#181) got merged into this branch instead of master, so this branch needs to get merged into master. There are no new changes in since that PR (except for merge commits).


MRichards99:
Recreated the original PR so this is no longer needed

PR #202, opened 2021-01-22
Add teardown for POST entity endpoint tests

### LINKED ISSUE #201: Fix No Teardown on Failed Create Data Tests
**Description**:
On the ICAT backend, if any of the tests prefixed with `test_valid` in `test_create_icat.py` fail, the data that gets created from those tests won't get removed, thereby not leaving icatdb into a clean state once the tests have finished running. This happens because the data is removed at the end of the test function (after the `assert` statement) so any code after that won't run if it throws an `AssertionError`.

**Acceptance criteria**:
- [ ] All tests should still pass
- [ ] If any tests in `test_create_icat.py` with the `test_valid` prefix are provoked to fail, the investigations table on `icatdb` should still be in the same state that it started. This can be indirectly checked through the other tests - if it hasn't been fixed, there'll be lots of other test failures where investigation data is involved

MRichards99:
This PR will close #201 

## Description
As per the linked issue, this fixes a bug found in `test_create_icat.py` where failing tests would prevent data from being deleted. The code for deleting test data has been moved to a pytest fixture, so will run regardless of the test result.

## Testing Instructions
To test, make `test_valid_create_data` and `test_valid_boundary_create_data` fail (change the response data perhaps so the assert statement won't be equal for example). Run `nox -p 3.6 -s tests -- test/icat/endpoints/test_create_icat.py` to just run that test class and see that they fail. Once failed, you need to verify the test data has been deleted. You can check `icatdb` using `mysql -p -D icatdb` and execute `SELECT * FROM INVESTIGATION ORDER BY ID DESC LIMIT 3;` or send a GET request to the API using filters to make an equivalent to that SQL query and there should be no test data in the ICAT instance. 

- [ ] Review code
- [ ] Ensure test data is deleted once the tests in `test_create_icat.py` is run (see above)

## Agile Board Tracking
Connect to #201

### DIFF
--- a/test/icat/conftest.py
+++ b/test/icat/conftest.py
@@ -12,6 +12,7 @@ from datagateway_api.src.api_start_utils import (
     create_api_endpoints,
     create_app_infrastructure,
 )
+from test.icat.endpoints.test_create_icat import TestICATCreateData
 from test.icat.test_query import prepare_icat_data_for_assertion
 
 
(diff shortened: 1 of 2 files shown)


patrick-austin (approved):
Seems to do everything it's supposed to, though I couldn't manage to directly check the state of the DB. But I did check the results of subsequent tests as the original issue suggests. Forcing a failure with `@pytest.mark.usefixtures("remove_test_created_investigation_data")` fails only the test in question, whereas commenting that line out and forcing a failure leads to the following test to fail as well.

PR #204, opened 2021-02-04
Bugfix/improve db sessions

louise-davies:
## Description
These are the changes to improve the session handling that have been on preprod for months. Essentially, the `session_manager` object was flawed, and reading the session docs for SQLAlchemy (https://docs.sqlalchemy.org/en/13/orm/contextual.html), and it suggests using an "off-the-shelf" solution in Flask-SQLAlchemy if we're unfamiliar with session handling/threading. I tried it as a quick fix, and it seems a lot more stable, so it's easier to just import Flask-SQLAlchemy purely for its session handling than to try and fix the original solution.

## Testing Instructions
- [ ] Review code
- [ ] Verify that preprod (which this branch is deployed on) is not having regular 500 errors due to sessions.


MRichards99 (approved):
I'm happy with these changes - I've tested them locally and on preprod and didn't experiences any issues on either.

PR #205, opened 2021-02-04
Make DB Backend use camelCase

### LINKED ISSUE #119: Casing of JSON fields is not consistent with ICAT API
Since ICAT is written in Java, it uses `camelCase` for it's JSON field names when querying entities. However, `datagateway-api` uses `SNAKE_CASE` since it reflects the database column names. This will cause incompatibilities when datagateway switches over from accessing this to the new ICAT API.

Acceptance criteria:
- [ ] Return objects match what ICAT returns w.r.t. casing

MRichards99:
This PR will close #119

## Description
The DB backend now uses camelCase, instead of SNAKE_CASE, to get closer to behaving like the ICAT backend. The majority of this PR consists of editing various types of names to match the field names shown in the ICAT schema (mostly found in `models.py`). For attribute names, this meant mapping variables in the entity model classes to the SQL columns (as per the code in the issue).  These name changes mean that the majority of them match what's found in the ICAT schema, but there is the odd difference with a couple of related entities (explained below). 

To make the change to camelCase for related entities, I've added `__singularfieldname__` and `__pluralfieldname__` for each entity - depending on whether there's a relationship involving a MANY, either singular or plural field names are used. Not all entities use both, but I've added them to all entities for the sake of defensiveness. These variables are used to ensure the related entity name is the same as the entity name uses as an input (in something like an include filter). The related entity name for inputs are defined by the `backref=` part of the SQLAlchemy relationships defined in `models.py`. There is the odd difference with `SampleParameter` (and the other types of parameter within the schema) because some entities use the field name of `parameters` to refer to them as a related entity, while others use `sampleParameters`.

I've moved `get_entity_object_from_name()` to a different file to prevent any circular import issues (the same reason the `endpoints` dictionary is now in a file on its own). Now this function doesn't live in the same file as the entity classes, `global()` has been replaced with `getattr()` which I think is a much cleaner solution - not sure why I didn't think of that at the time... The function also now accepts inputs of a plural entity name - this feature is utilised when user input of an include filter contains a plural field name.

Unit tests, the `GET` session details, ISIS endpoints, `modTime` etc. attributes have all been adjusted to match the camelCasing changes of this PR.

I've added appropriate file ignores in `.flake8` to ensure the camelCase doesn't interfere with flake8's linting. I've also edited the `.gitignore` for the log file as I found rotated log files were appearing in my modified file list from Git. A rebuilt `openapi.yaml` file is also contained within these changes (alongside an adjusted function that deals with creating the schema models for that file in `entity_map.py`).

## Testing Instructions
I didn't add any additional testing for these changes (just fixed the existing tests), but we should discuss (as a team) whether having tests that test every single entity on every single request is needed or not (as opposed to just testing each type of request using a single entity, which is mainly investigation) or if releasing the API is of higher priority.

- [x] Review code
- [x] Output of GET & PATCH requests uses camelCase keys
- [x] POST requests accept camelCase key inputs
- [x] Include filter accepts camelCase inputs (e.g. `userGroups`)
- [x] Where filter accepts inputs using included/related entity (e.g. `{"userGroups.id": {"lt": 10}}` with an appropriate include filter)

## Agile Board Tracking
Connect to #119


MRichards99:
That's odd, all the tests pass locally, I shall investigate...

MRichards99:
Based on my work on Tuesday (which will continue today) I can't guarantee this is the final set of changes for this issue, so I've converted it to a draft for the time being. I will convert it back to 'ready to review' once I can verify the changes on this branch can be applied to the frontend.

MRichards99:
Now the camelCase work on the frontend is coming to an end, here's an update on this PR on further changes that have been made:
- Fixed a bug where a request contains a WHERE filter using an `in` operator containing an empty list resulting in a 500. DB backend gave back a 200, so I manipulated the Python ICAT backend to give the same result. Commit: 6662f86bff60397140d77d996b95ae981bc5073b
- Where no data is found from a query such as a get with filters (ICAT backend), a 404 used to be returned. This is incorrect as no data is the correct response. An empty list (and therefore a 200) is returned instead. Commit: c1fc105446aebcd9b393ec9d2135fe6fad49f791, 268dac3b2e90b5caa0fce66fb470eac96b63d565
- Foreign keys now follow the form of `facilityID`, rather than `facility`. This change means that the DB backend would be compatible with DataGateway if we so desired to do that again, where this wasn't the case earlier on in this branch because of foreign keys and related entities using the same names. Commit: 85b76b6c7da691cd9059f304554d0d59104b93a5, 7912f9628cf7eebfbca3a01439fdf70c8d88a052, 1ee79619f0a97cf4dc9c72a8f0979c44bceca750

I foresee no further work required on this branch to get the frontend working on the ICAT backend, so this is ready to review.

MRichards99:
The dependency safety part of the CI has failed, however this is fixed in a commit (622eebb6ac8009154dd957d4919854cfe3eefaaf) on a different branch (currently in PR). Once that PR is merged (and master merged back into this branch), this will pass.

datagateway_api/common/database/models.py
```diff
@@ -170,7 +153,7 @@ def get_related_entity(self, entity):
         :return: The entity
         """
         try:
-            return getattr(self, entity)
+            return getattr(self, entity if entity[-1] == "s" else entity.upper())
```
louise-davies, line 156: Is this testing for "pluralness" - it works fine for now, but perhaps a comment explaining this would be helpful in case an new ICAT entity type is created which doesn't "pluralise" with an "s" on the end.

datagateway_api/src/resources/entities/entity_map.py
```diff
@@ -106,17 +68,25 @@ def create_entity_models():
...
-                    "$ref": f"#/components/schemas/{relationship_name.strip('_')}",
+                    "$ref": "#/components/schemas/"
+                    f"{relationship_name.strip('_').upper()}",
                 }
             if (
                 relationship_class.direction.name == "MANYTOMANY"
                 or relationship_class.direction.name == "ONETOMANY"
             ):
+                entity_underscore_strip = relationship_name.strip("_")
+                if entity_underscore_strip[-1] == "s":
```
louise-davies, line 79: Again, probably useful to specify in a comment that this is checking for plurals

louise-davies (commented):
Looks good - just some ideas for comments for future proofing

datagateway_api/common/database/models.py
MRichards99: Updated the docstring of this function to give some detail about the logic behind the plural stuff.

datagateway_api/src/resources/entities/entity_map.py
MRichards99: As above

PR #216, opened 2021-04-07
Improve Client Handling

### LINKED ISSUE #209: Implement Client Reuse/Caching
**Description**:
Currently, the API creates a new `client` object from Python ICAT every time one is needed (for every endpoint that requires a session ID in the header and the `/sessions` endpoints). This makes the API sluggish, with Postman measuring response times in seconds, rather than ms. In turn, this has caused issues with end-to-end tests on DataGateway (particularly on the work for https://github.com/ral-facilities/datagateway/issues/94).

I've created a proof of concept for client reusing (by creating a single client object at start-up and passing that around using `kwargs`) which has been successful for my local instance (e2e tests on DataGateway are happier!) but won't be robust for production. Some form of client caching (using LRU caching) is needed, though there is an edge case with endpoints that need the client object but have no session ID at the point of needing it (e.g. POST /sessions).

**Acceptance criteria**:
- [ ] Multiple users should be able to use the API without getting logged out etc.
- [ ] GET requests should take a more sensible amount of time (not 12-ish seconds for a basic count request). This should probably be measured by using DataGateway e2e tests - these should pass without needing to add lots of timeouts (I was adding timeouts of between 100 and 200 seconds before my proof of concept)
- [ ] API functionality shouldn't be affected - i.e. unit tests should pass

MRichards99:
This PR will close #209 

## Description
This branch improves how client objects are managed across the API. Before this change, client objects were created towards the beginning of each request. These objects take several seconds to create, so caused the API to run very slowly. Clients are now managed by an LRU cache, which fetches clients from an object pool when needed. A pool of clients is created at startup, and are moved between the cache and back to the pool as required. I've documented how this works, so hopefully what I've put in there (and comments in the code) should be enough to not need to repeat all that in this PR. 

For reference, 74f1edd78b56b10618ec1c57f8016eb6b5073ad7 is the first commit where the cache-pool design appears and the commit message gives quite a bit of detail.

I've added tests for the LRU cache class that I've extended to ensure it's created with the configured cache size and that popping a client off the cache releases it back to the pool.

## Testing Instructions
Follow the documentation (https://github.com/ral-facilities/datagateway-api/blob/bugfix/client-cache-%23209/README.md#client-handling) to configure the cache and pool sizes. The login endpoint (`POST /sessions`) can be used to create session IDs, and any of the GET endpoints can be used to allocate a client object to that session ID. The combination of those two actions will allow the reviewer to test the cache and pool design, to ensure it works as intended when configured to correct proportions.

- [ ] Review code
- [ ] Ensure the cache can grow to the configured size and that popping a client off the cache releases it back to the pool
- [ ] Ensure the pool creates the configured number of objects at startup (will be printed out to the console)
- [ ] Ensure the pool can grow over the initial size
- [ ] Test the e2e tests on DataGateway are happy with this change. My changes on `bugfix/casing-json-fields#94` aren't complete (some e2e tests fail), but running the DataGateway Download e2e tests should give a good representation (they pass happily on my machine using this branch)

## Agile Board Tracking
Connect to #209


VKTB (dismissed):
The code changes look good to me. Someone with a better knowledge of client handling and the API itself should have another look in case I missed something, and perform the manual tests that you listed at the top.

stuartpullinger (approved):

PR #219, opened 2021-04-14
Fix Distinct Filter on ICAT Backend

### LINKED ISSUE #141: implement distinct filter in python-icat
pass through list of fields to icat. May need to transform the result returned from ICAT to only include the requested fields.

MRichards99:
This PR will close #141 

## Description
These changes make the distinct filter work on the ICAT backend, returning actual distinct results, not results that are repeated (the previous behaviour on this backend). 

The previous implementation used WHERE filters (filtering on columns as per the distinct filter with `!= null`) to filter out empty columns. The results were returned from Python ICAT and the API picked out the columns requested on the distinct filter. The new implementation uses the `setAttributes()` function in Python ICAT to select only the columns requested in the distinct filter (similar to `SELECT colA, colB, colC` in SQL). The `DISTINCT` keyword is also applied to the query.

When a distinct filter is used in a GET request (except for a count request), `entity_to_dict()` is not used, but a new function has been created to process the results from Python ICAT to a format suitable to be returned to the user (called `map_distinct_attributes_to_results()`). `entity_to_dict()` assumes a data input of an ICAT entity object, a format which is not returned by Python ICAT when `setAttributes()` is used in the query (a list is returned instead).

Two boolean flags have been created for the query, so the query class can deal with edge cases:
- `manual_count` - this flag is used when a `DISTINCT` keyword is used on a count endpoint (so a distinct filter on a count request, but also the count ISIS endpoints). JPQL cannot handle the `DISTINCT` and `COUNT` keywords in the same query, so a normal (non-count) distinct query is sent to ICAT and the API performs `len()` on the result to get the count. I've written a comment in `abb31bd0c36a809146484a78c8c6d9864fdf561a` so future developers understand why this is required
- `isis_endpoint` - this flag is used to distinguish ISIS endpoints from standard endpoints. This is required as these endpoints use the `DISTINCT` keyword but don't select multiple attributes. This flag is used to divert the results to `entity_to_dict()` (correct function to be used) instead of `map_distinct_attributes_to_results()` (incorrect). Commit `504e94feef9a608796197e1ae2ac9b67019ec843`.

I've removed some logic that checks if related entities on a distinct filter have an associated include filter as I've realised Python ICAT is smart enough to include JOINs in the JPQL queries it constructs so mandating this is unnecessary really (`1530ca69b3f892349eeb2ba4ed9df94cce441d34).

I've removed the old distinct field functions at `0322a99c4738a356cbec89f65eaccf5a89c918fc` and `677424d8bdf80d65ea6d8ce425f33fde995ee96d`.

I've also added some test cases to test all these changes. Tests that use the flask test app, but also proper unit tests which test a single function. And of course I've removed the tests for the functions that are no longer used.

## Testing Instructions
The API now uses 0.18.0 of Python ICAT which implements changes in being able to select multiple attributes (required for this work). For the distinct filter, 0.18.1 of Python ICAT is required, but isn't currently released. As a result, some unit tests (3 I think?) will fail on CI until it gets released. For the time being, I have manually installed Python ICAT using the GitHub branch for the upcoming release (`pip install git+https://github.com/icatproject/python-icat.git@release/0.18.1`). Pip might complain about version `UNKNOWN` not matching `0.18.0` (as mandated by Poetry) but ignore this, the installation will have been applied. When 0.18.1 is released, I will update `pyproject.toml` to reflect this.

Distinct filters should be able to get applied on GET requests. Applying a distinct filter on a count endpoint should also have an impact the number of results returned. These cases are covered in `test/` but a good manual test case can be found [in a comment on the original issue](https://github.com/ral-facilities/datagateway-api/issues/141#issuecomment-728812545).

- [x] Review code
- [x] Ensure distinct filter behaviour works as expected

## Agile Board Tracking
Connect to #141


MRichards99:
0.18.1 of Python ICAT has been released so I've updated the API's dependencies to reflect that. It should be noted, ICAT Server 4.11.1 is required as a result.

MRichards99:
@louise-davies 

As mentioned in sprint planning yesterday, when you have a DISTINCT filter with an include filter, there are issues. The only case where this is actually used in the frontend is on card view, where you can filter by type and facility ID. I've fixed this by just removing the include filter on those requests (see: https://github.com/ral-facilities/datagateway/commit/b2374ceaea49e84b36cd60c6b1efab8e827a914e) - Python ICAT is smart enough to automatically add a JOIN to the JPQL query.

I've just been using `/entityManager` to investigate if it's a bug on the API side, or if it's a JPQL limitation. It is a JPQL limitation as proven by the following query:
```sql
SELECT DISTINCT o.fullName, o.email FROM User o INCLUDE o.userGroups
```

The entity manager will return:
```json
{
    "code": "BAD_PARAMETER",
    "message": "INCLUDE is only valid for one quantity in the SELECT clause"
}
```


An alternative test query:
```sql
SELECT DISTINCT o.fullName FROM User o INCLUDE o.userGroups
```

Result:
```json
{
    "code": "BAD_PARAMETER",
    "message": "variable O mentioned in INCLUDE clause is not defined"
}
```
I'm happy this is a JPQL limitation so please feel free to start reviewing this PR, I have no further pre-review changes to make.

louise-davies (approved):
Looks good to me - I took the liberty of merging in master to run the CI :)

PR #222, opened 2021-04-20
Refactor config to be aware of the backend in use

### LINKED ISSUE #210: Make missing config errors backend specific
**Description**:
Currently, you need to specify DB_URL even if you're using the ICAT backend and presume similarly for ICAT config values when using the DB backend. It's not a game changer - it's just a little annoying to specify an empty string for a value you don't even care about!

**Acceptance criteria**:
- [ ] Config items take into account the backend before throwing an error

MRichards99:
This PR will close #210 

## Description
This is a refactoring for the API's configuration. The root issue described in #210 was caused by calls to config getters in `constants.py` - as this file was imported on startup, there was a requirement for `DB_URL` and `ICAT_URL` to be present. 79d00e8b61509f5dc92b008057d9b3bb91ead3b2 shows the fix for this root issue, with the commit message giving some further explanation. 

3c40ccb546f0434d82366d9a521ac625e6020220 & 9ad2f574ff461e66a59ad0ea2b072d71bf78533b - I've replaced each of the getter functions with a single, generic getter function, taking inspiration from SciGateway Auth on this. Unlike Auth, the config is not hot reloaded as there's no requirement for the API to have this functionality.

Beforehand, the config options included `ICAT_URL` and `DB_URL` - I've changed these to lowercase snake_case to match the formatting of the others. c1a02e5e0aac93906bf4589a17b51657d59c7163 for that change. 

cab681b5e663bcd214aa4f386c45928f8e9a5390 - a function has been added to check that all relevant config options are present in `config.json` at startup. This looks at the backend set and determines which of the backend specific config options must be present for all API functionality to work.

171dbdfee702e8760bd4e130c0738a29675b8618 & 76bcb7d783b8c8f52978a044863e764e63111b2b - I took some further inspiration from SciGateway Auth and implemented an enum class for the API's config options - as the docstring indicates, this is more for ease of use for development rather than bringing any real functional improvements.

## Testing Instructions
I've changed `test_config.py` to reflect these changes, but for manual testing, you should try launching the API with all the config options which startup the API with no issues. Then try removing config option(s) that aren't backend specific, the API should exit when it finds the missing config option, outputting what's missing in the terminal. Do the same for each backend (and their backend specific options), but first set the backend to whichever one you're removing config options from.

- [x] Review code
- [x] Test that removing config options are spotted at startup, for non-backend specific, ICAT specific and DB specific

## Agile Board Tracking
Connect to #210


test/icat/test_session_handling.py
```diff
@@ -90,6 +101,7 @@ def test_refresh_session(self, valid_icat_credentials_header, flask_test_app_ica
     def test_valid_login(
         self, flask_test_app_icat, icat_client, icat_query, request_body,
     ):
+        print(request_body)
```
VKTB, line 104:
```suggestion
```

test/test_config.py
```diff
@@ -3,159 +3,58 @@
...
+        [
+            pytest.param("python_icat", id="Python ICAT Backend"),
+            pytest.param("db", id="Database Backend"),
+        ],
+    )
+    def test_valid_config_items_exist(self, backend_type):
+        test_config = Config(
+            path=Path(__file__).parent.parent
+            / "datagateway_api"
+            / "config.json.example",
```
VKTB, line 37: Can't we use the `test_config` fixture at the top of this file to do this?

datagateway_api/common/config.py
```diff
@@ -9,17 +10,84 @@
...
+        :param config_key: Enum of a configuration key that's in `config.json`
+        :type config_key: :class:`APIConfigOptions`
+        :return: Config value of the given key
+        """
         try:
-            return self.config["backend"]
+            return self._config[config_key.value]
         except KeyError:
-            sys.exit("Missing config value, backend")
+            sys.exit(f"Missing config value: {config_key}")
```
VKTB, line 90: This prints `Missing config value: APIConfigOptions.ICAT_URL` and I think that it would be better if it prints the actual name of the config key. You can get the key name by doing the following:
```suggestion
            sys.exit(f"Missing config value: {config_key.value}")
```

VKTB (changes requested):
Looks good just a couple of minor things.

test/test_config.py
MRichards99: Good point! This test function now uses the fixture.

VKTB:
@MRichards99 Latest changes look good to me. The conflicts will need to be resolved before I hit the approve button.

MRichards99:
The conflicts must've occurred when I merged the other PR into master. I've resolved the conflicts now and got everything working :)

VKTB (approved):

PR #232, opened 2021-05-05
Add Timezone Data to DB Backend

### LINKED ISSUE #225: Use DateHandler when dealing with dates on DB backend
**Description**:
The ICAT backend uses a datetime format which includes timezone data (`2000-05-07 00:00:00+00:00`). This is managed by the `DateHandler` class (stored in `datagateway_api/common/date_handler.py`).

This is part of the work for #213 

**Acceptance criteria**:
- [ ] Both backends should have timezones and follow the same datetime format (as above)
- [ ] A user should be able to easily 'round trip' data involving datetimes (i.e. copy from a GET, paste into a PATCH request without needing to change the format) as this is possible on the ICAT backend

MRichards99:
This PR will close #225 

## Description
This change helps to make the DB backend become more aligned with the ICAT backend by having timezones when displaying datetimes on the API. 

It seems the DB backend doesn't actually use timezones anywhere - I tried setting the `timezone` flag as pointed out in the [docs](https://docs.sqlalchemy.org/en/14/core/type_basics.html#sqlalchemy.types.DateTime) but that didn't have any impact. Instead, I've made the DB backend display the local timezone for the API. This seems to be the best compromise I could find to make the two backends aligned.

The DB backend now uses the `DateHandler` so will have the same format for datetimes as the ICAT backend.

I also noticed that the get session details endpoint doesn't use timezones on the ICAT backend. I fixed that on 9b2dc7d.

## Testing Instructions
Check that both backends have display datetime in the same format (including the timezone part of the output)

- [ ] Review code

## Agile Board Tracking
Connect to #225

### DIFF
--- a/datagateway_api/common/constants.py
+++ b/datagateway_api/common/constants.py
@@ -1,6 +1,8 @@
 from datetime import datetime
 
+from dateutil.tz import tzlocal
+
 
 class Constants:
     PYTHON_ICAT_DISTNCT_CONDITION = "!= null"
-    TEST_MOD_CREATE_DATETIME = datetime(2000, 1, 1)
+    TEST_MOD_CREATE_DATETIME = datetime(2000, 1, 1, tzinfo=tzlocal())
--- a/datagateway_api/common/database/models.py
+++ b/datagateway_api/common/database/models.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from decimal import Decimal
 import enum
 
+from dateutil.tz import tzlocal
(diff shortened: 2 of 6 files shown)


sam-glendenning (approved):
Works fine. Example output with a GET request to `/sessions`:

```
{
    "id": "e5525746-b235-11eb-bbcb-000c2906e44d",
    "expireDateTime": "2021-05-12 01:50:11-07:00",
    "username": "simple/root"
}
```

Issue #227, opened 2021-04-26
Fix incorrect related entity references


MRichards99:
**Description**:
Last week, I noticed some of the foreign key references don't match the ICAT schema on DB backend. In particular, `type` (related entity from investigation to investigation type) could only be referenced using `INVESTIGATIONTYPE` (it seemed to be case insensitive actually). These foreign key references should be checked and modified to match the ICAT schema.

This is part of the work for #213 

**Acceptance criteria**:
- [ ] Related entity references should match ICAT schema (and therefore ICAT backend) - this is most easily tested with include filters on GET requests of the different entities

MRichards99:
Unassigning myself as this won't be tackled in the short term.

MRichards99:
Adding the wontfix label to this issue - there's no plans to work on this issue for the time being and because the DB backend won't be used in production, this is a low priority, potentially high effort issue.

PR #233, opened 2021-05-17
Allow related entities on DB distinct filter

### LINKED ISSUE #223: Allow distinct filter to accept inputs with related entities
**Description**:
On the ICAT backend, the following request is acceptable and retrieves distinct investigation type IDs:
`http://{{datagateway-api}}/investigations?distinct="type.id"`

On the DB backend, this doesn't work as the distinct filter doesn't recognise the input contains a related entity, so a 400 is returned to the user. As part of the work for #213, the distinct filter on the DB backend should recognise a related entity in its input (perhaps by making use of `.split(".")` and adding a JOIN to the query?) so that equivalent behaviour works on both backends. 

The use of related entities is used in the move to the ICAT backend on DataGateway.

**Acceptance criteria**:
- [ ] DB backend GET requests (i.e. where distinct filters are acceptable to use), should accept distinct filters containing a related entity reference and return the correct result (`type` or `investigationtype` for the DB backend currently)

MRichards99:
This PR will close #223 

## Description
This branch allows related entities to be used on the distinct filter on the DB backend. The tests I've added in `test_get_with_filters.py` show the additional functionality this branch brings.

I've moved a lot of the functionality that was already present in `filters.py` into another class, `DatabaseFilterUtilities` - I've done this so both the where filter and distinct filter can make use of the same code. (1b7f1325c433cf0917bdbc8e557468a1f110296e contains the bulk of the class creation, 11f75a20a2dba6aac446fbaeb8d61a3f8a436233 removes most of the pre-existing, non-generic code). I've added a test class to test this class (bc9635153bae681e89962fb37d71c14cf16056e7).

I've separated out functionality of `add_query_join()`, more info in the commit message of a2538ccba6810fbd23c6f2c9d1d09ea89ba8498b

This morning, I properly fixed a bug I attempted to fix on Friday (proper fix at 9021fc865872586441fe710b29f7cfbd9ca87522). I tried to use a flag to limit when `add_query_join()` was called, but that didn't work in cases where there was related distinct fields and a where filter in the same request. f16cb15a4360e6fb818db8409508d6f9352127e0 shows actual cases that this bug fixes.

## Testing Instructions
I've added test cases to test/demonstrate the functionality added here. The main thing is that the functionality added here matches what's available in the ICAT backend so construct some GET requests with distinct fields (mixture of related and unrelated) and check they give the same results. You might find that some entity names need to be referenced differently in DB backend (e.g. `investigationType` which should be `type` according to the ICAT schema) but I'm aiming to fix those issues in #227.

- [ ] Review code
- [ ] Check the distinct functionality is the same in both backends

## Agile Board Tracking
Connect to #223

### DIFF
--- a/datagateway_api/common/database/filters.py
+++ b/datagateway_api/common/database/filters.py
@@ -2,7 +2,6 @@ import logging
 
 from sqlalchemy import asc, desc
 
-from datagateway_api.common.database import models
 from datagateway_api.common.exceptions import FilterError, MultipleIncludeError
 from datagateway_api.common.filters import (
     DistinctFieldFilter,
--- a/datagateway_api/common/database/filters.py
+++ b/datagateway_api/common/database/filters.py
@@ -40,7 +40,6 @@ class DatabaseFilterUtilities:
         self.field = None
         self.related_field = None
         self.related_related_field = None
-        self.distinct_join_flag = False
 
     def _extract_filter_fields(self, field):
         """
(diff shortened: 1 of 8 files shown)


sam-glendenning (approved):
Works well for me. Tried it with:

- `/investigations?distinct="type.id"`
- `/applications?distinct="id"`
- `/datacollections?distinct="modId"`

and a couple of others and got the same results (not including any differences in entity names that you've said is being fixed in #227).

Issue #224, opened 2021-04-26
Don't display foreign keys in data output


MRichards99:
**Description**:
In the DB backend, foreign keys are returned in the response body of a GET request. Example for ``http://{{datagateway-api}}/datasets?limit=1&include="sample"``: 

```json
[
    {
        "id": 1,
        "complete": false,
        "createId": "user",
        "createTime": "2011-01-29 06:19:43",
        "description": "Long bed after near eye wrong. Develop Mr everybody common.\nDemocrat until stock southern song. Close face test which summer.",
        "doi": "0-467-23440-X",
        "endDate": null,
        "location": "/million/young.mp3",
        "modId": "user",
        "modTime": "2008-10-15 12:05:09",
        "name": "DATASET 1",
        "startDate": "2000-05-07 00:00:00",
        "investigationID": 1,
        "sampleID": 1,
        "typeID": 1,
        "sample": {
            "id": 1,
            "createId": "user",
            "createTime": "2011-01-29 06:19:43",
            "modId": "user",
            "modTime": "2008-10-15 12:05:09",
            "name": "SAMPLE 6",
            "investigationID": 6,
            "sampleTypeID": 10
        }
    }
]
```

The foreign keys shouldn't be displayed on output as they're not displayed on the ICAT backend - this is part of the work for #213.

As a note, there could be some cases where the helper functions that fetch results from the database are used as an input into another function (updating data perhaps). The ICAT backend uses similar design, so a flag might be required to distinguish the two use cases (no foreign keys GETs and foreign keys for DB input). Not 100% sure on this, but something to keep in mind.

**Acceptance criteria**:
- [ ] Foreign keys shouldn't be displayed GET, POST or PATCH requests involving entities that have relationships with other entities

MRichards99:
Unassigning myself as this won't be tackled in the short term.

MRichards99:
Adding the wontfix label to this issue - there's no plans to work on this issue for the time being and because the DB backend won't be used in production, this is a low priority, potentially high effort issue.

PR #235, opened 2021-05-21
Fix complex include filters

### LINKED ISSUE #234: Fix error with complex include filters
**Description**:
While merging DataGateway branches today, I've noticed a bug on the `PythonICATIncludeFilter` on complex filter inputs, such as in the following request:

`http://{datagateway-api}/studyinvestigations?order="id+asc"&include=["study",{"investigation":[{"investigationUsers":"user"},{"investigationInstruments":"instrument"}]}]`

**Acceptance criteria**:
- [ ] A checklist
- [ ] of criteria to accept this story as done

MRichards99:
This PR will close #234 

## Description
As per the issue, a complex include filter input didn't work in the ICAT backend, but did work in the DB backend. This is a small change to fix that, with a couple of tests added to demonstrate this.

## Testing Instructions
Check the request in the issue (and ones similar to it) now work and don't return errors.

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Connect to #234

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -11,4 +11,5 @@ per-file-ignores =
   util/icat_db_generator.py: S311
   datagateway_api/wsgi.py:E402,F401
   datagateway_api/common/database/models.py: N815,A003
+  datagateway_api/common/icat/filters.py: C901
 enable-extensions=G
--- a/datagateway_api/common/icat/filters.py
+++ b/datagateway_api/common/icat/filters.py
@@ -246,7 +246,9 @@ class PythonICATIncludeFilter(IncludeFilter):
                             self._extract_filter_fields(".".join((key, element)))
                         elif isinstance(element, list):
                             for sub_element in element:
-                                self._extract_filter_fields(".".join(key, sub_element))
+                                self._extract_filter_fields(
+                                    ".".join((key, sub_element)),
+                                )
                         elif isinstance(element, dict):
                             for (
                                 inner_element_key,
--- a/datagateway_api/common/icat/filters.py
+++ b/datagateway_api/common/icat/filters.py
@@ -241,8 +241,29 @@ class PythonICATIncludeFilter(IncludeFilter):
                     self._extract_filter_fields(".".join((key, value)))
                 elif isinstance(value, list):
                     for element in value:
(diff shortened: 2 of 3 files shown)


sam-glendenning (commented):
So I tried that specific query out and I seemed to get different results. Both give a lot of output but I put it into a JSON diff tool and got the below (`db` backend on the left, `python_icat` on the right):

diff

There's many more differences beyond this image, this is just a snapshot. Happy to demonstrate this over Zoom but I'd like to see what anyone else got in case I'm missing something

MRichards99:
So the foreign keys on the left (e.g. `facilityID`) will hopefully be removed in #224. `pid` inside `instrument` is there because the DB backend is based on a version of ICAT Server older than 4.10.0 (which is around then that attribute was introduced and the same thing happens with `user.affiliation`. I'm not sure the exact version the DB backend is based on, but 4.8.0 or 4.9.0 would be my guess. Should've put those details in the PR but I was trying to get it done by the end of Friday afternoon :) I think those are the only differences, unless I've missed some others?

I've just pushed a commit which fixes a bug I introduced on this branch - just a small mistake with `join()` - I've added a test case to monitor that too.

sam-glendenning:
> So the foreign keys on the left (e.g. `facilityID`) will hopefully be removed in #224. `pid` inside `instrument` is there because the DB backend is based on a version of ICAT Server older than 4.10.0 (which is around then that attribute was introduced and the same thing happens with `user.affiliation`. I'm not sure the exact version the DB backend is based on, but 4.8.0 or 4.9.0 would be my guess. Should've put those details in the PR but I was trying to get it done by the end of Friday afternoon :) I think those are the only differences, unless I've missed some others?
> 
> I've just pushed a commit which fixes a bug I introduced on this branch - just a small mistake with `join()` - I've added a test case to monitor that too.

@MRichards99 As long as the differences I've mentioned can be accounted for, I'm happy to approve it. Although your codecov isn't happy...

MRichards99:
Yeah I can't see anything else that concerns me. Overall codecov diff is -0.01% now so pretty marginal

sam-glendenning (approved):

PR #237, opened 2021-06-04
Make Generator Script Consistent

### LINKED ISSUE #236: Make icatdb generator script generate identical data
**Description**:
The generator script stored in `util/` generates different data each time it runs. From my testing, IDs of data get jumbled up and data generated using the standard random library (so `choice()` and `randrange()` is different each time.

Once these fixes have been made, the preprod database should probably get dropped/re-created and the script re-run because the data currently on that database is different enough that the current seed will still produce different data (compared to preprod) once these issues have been fixed. This view is based on some diffs I've been doing today using JSON files I'm happy to share and explain if requested. The e2e tests should probably be 'fixed' to reflect the new data before this happens though (shall create an issue for this).

**Acceptance criteria**:
- [ ] Data generated from this script should be identical each time - I've just been testing this on a few entities (investigation, user, parametertype) and getting the first 30 results via the API
- [ ] The data generated from this script should agree with DataGateway - i.e. the e2e tests should be able to pass on that repo once the relevant changes have been made

MRichards99:
This PR will close #236 

## Description
This PR mean `icat_db_generator.py` will now generate consistent data across multiple uses of it. The data generated is different to the current data on preprod, so that will need generated (once I've fixed DataGateway's e2e tests to match the new data), but that's monitored elsewhere. 

There were two main causes for the inconsistency - the `multiprocessing` and `random` libraries.

Using the multiprocessing pool to map the generator functions to the number of data points needed was causing the IDs to be jumbled up - I guess the data was getting generated and inserted into the database in different orders each time. I've removed this pool (556faa0c2dcd9a88cbbf09196c3df76f5464cc13) which does make the script run slower, but I think consistency > speed in this use case. Perhaps the pool and execution of `post_entity()` could've been refactored to allow the multiprocessing aspect of the script to be kept in, but I'm not sure it's worth the time sink for now.

Despite using the `random` library's `seed()` function, it didn't really seem to help for data (integers) being generated. I've removed usage of that library and replaced it with equivalents from Faker (327557ff59de14954c13d1a66cea99118d478d98 and 7671a7d8acb78f791e7f26464ace2823e4fbb2fc) which does produce consistent integers - this includes the previous use of `random.choice()`.

I've also updated the version of Faker (ede12bf49be374f6655a3339c53dfd0834df153c), in fact, to one that was released yesterday!

While working on this, I've been doing manual testing, comparing API outputs from the first 30 results on a few entities, though I think this should be kept track of from now on. Ideally, I would've liked to have added something to live-compare preprod's data against what the script produces, but preprod's API port will be closed off in the future so won't be possible to do beyond then.

Instead, I've added a job on the GitHub Actions workflow to run the generator script, produce a dump, clear the database, repeat the generator script and diff the dumps. Hopefully this shows the script now does produce consistent results, though I'm happy to share my manual testing process if needed.

## Testing Instructions
Run the script and ensure it runs correctly

- [ ] Review code

## Agile Board Tracking
Connect to #236

### DIFF
--- a/util/icat_db_generator.py
+++ b/util/icat_db_generator.py
@@ -2,7 +2,6 @@ from abc import ABC, abstractmethod
 import argparse
 import datetime
 from multiprocessing import Process
-from random import choice, randrange, seed
 
 from faker import Faker
 from sqlalchemy import create_engine
@@ -34,7 +33,6 @@ SEED = args.seed
 YEARS = args.years  # 4 Cycles per years generated
 faker = Faker()
 faker.seed(SEED)
-seed(a=SEED)
 
 
 engine = create_engine(
@@ -328,7 +326,9 @@ class InvestigationGenerator(Generator):
         investigation.summary = faker.text()
         investigation.title = faker.text()
         investigation.visitId = faker.random_int(1, 100)
-        investigation.typeID = randrange(1, 4)
+        investigation.typeID = faker.random_int(
+            1, InvestigationTypeGenerator.amount - 1,
+        )
         post_entity(investigation)
 
 
@@ -346,7 +346,7 @@ class InvestigationUserGenerator(Generator):
         apply_common_attributes(investigation_user, i)
         investigation_user.role = ["PI", "CI"][faker.random_int(0, 1)]
         investigation_user.investigationID = i
-        investigation_user.userID = randrange(1, UserGenerator.amount)
+        investigation_user.userID = faker.random_int(1, UserGenerator.amount - 1)
         post_entity(investigation_user)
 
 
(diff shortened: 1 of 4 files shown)


util/icat_db_generator.py
```diff
@@ -507,16 +526,18 @@ class InvestigationParameterGenerator(Generator):
     amount = InvestigationGenerator.amount
 
     def generate(self):
-        self.pool_map(generate_investigation_parameter)
+        for i in range(1, self.amount):
+            generate_investigation_parameter(i)
 
 
 def generate_investigation_parameter(i):
+    # TODO - Why's this not in the class?
```
sam-glendenning, line 534: Beats me. I added it back in as a static method and didn't see any difference in generated data. Wanna add it back?

sam-glendenning (approved):
Nice work here, @MRichards99 this works very well. Script runs slightly slower than before as you mentioned (~5 mins) but is a good trade-off for consistent data. One step closer for the local CI changes on DataGateway too :)

util/icat_db_generator.py
MRichards99: I forgot to search TODO in VS Code! The inconsistency irked me slightly (it makes no actual difference, but why's the function put in a different place to all the others), I shall resolve now

MRichards99: Change made and TODO removed :)

sam-glendenning (approved):

PR #240, opened 2021-07-19
Allow One-Many Related Orderings to not miss out blank results

### LINKED ISSUE #238: Implement Join Specs for Ordering on 1-many Relationships
**Description**:
Implement the changes brought by https://github.com/icatproject/python-icat/pull/85 on order filters that have a one-to-many relationship. This will fix the sorting issues discussed in https://github.com/ral-facilities/datagateway/issues/460

**Acceptance criteria**:
- [ ] Sorting DOIs on ISIS investigation table view on DataGateway should work correctly (results should be visible even if there's no DOI)

MRichards99:
This PR will close #238 

## Description
These changes fix sorting issues that have been raised in https://github.com/ral-facilities/datagateway/issues/460. This has been done by implementing a new feature in Python ICAT which allows us to specify which JOIN we'd like to use for particular JOINs. In my testing for Rolf, I found that LEFT JOINs (on the entities with a one-many relationship) solved the issues in DataGateway so this has been added in a programmatic way for DataGateway API. I've added unit tests which simulate a DataGateway use case and I've tested this on DataGateway (screenshot from ISIS dev data) to prove it has the intended effect:
image

I also did further testing and ran all the e2e tests (on `use-local-ICAT-in-CI-runs-#641` as my `icatdb` is set for this branch) to see if there were any unintended side effects. All tests passed and in my manual testing I couldn't find anything. 

At the time of creation, I'll leave this as a draft PR as I'm unable to update Python ICAT's dependency version since the new feature hasn't been released. It's possible to test this by manually uninstalling/reinstalling Python ICAT on the `join_specs` branch (almost bypassing Poetry for that) but it's probably just easier to wait for the new release as there's not any mass urgency I don't think.

## Testing Instructions

- [ ] Review code
- [ ] Check sorting on DOIs on ISIS investigation table view is all good (as per screenshot)

## Agile Board Tracking
Connect to #238


MRichards99:
Rolf has released Python ICAT 0.19.0 so I've updated that in the poetry files and sorted out the unit test which was failing on CI.

sam-glendenning (approved):
Works! Tested on a few tables containing an instrument column including facility cycle investigation table, study hierarchy investigation table and investigation search table. Here's an example with facility cycle investigations:

image

MRichards99:
Only just realised this was reviewed, Friday was a hectic day :)

PR #247, opened 2021-08-11
Add CI Generator Script Comparison

### LINKED ISSUE #246: Generator Script Data Branch Comparison
**Description**:
Now the change of preprod's generated data is coming to an end, the API's CI should test/compare the data on preprod against that generated on the branch the CI is being run against. The preprod SQL dump is ~15MB so it seems inappropriate to weigh the repo down with that (by storing it in `util/`, so I think an alternative approach is required.

One solution would be to compare the generated data from the generator script on master, against the generated data from the script on the branch the CI workflow is being run on. I'll add this to the end of the existing icatdb CI job.

**Acceptance criteria**:
- [ ] The existing generator script CI job (which tests that the script's seeding works) should also compare dumps between master and the branch the workflow is being run against

MRichards99:
This PR will close #246 

## Description
This is a small addition to the generator script CI job to compare generated data between master and the branch the CI is being run against. This will ensure that any future differences of the generator script are shown here rather than being hidden until we find them one day on preprod. 

As per the issue, I did think about storing the actual preprod dump on the repo (or maybe even the icatproject website?) but I decided against this due to the dump's size and I wanted to prevent any future schema changes causing differences between dumps (even if the data remains the same).

I've done a diff against the preprod dump and a dump of my local instance and they're the same, so we're starting from an 'identical base' to make this test actually meaningful. 

## Testing Instructions
I guess just check the CI steps make sense?

- [ ] Review code

## Agile Board Tracking
Connect to #246

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,6 +8,9 @@ Add a set up instructions describing how the reviewer should test the code
 
 - [ ] Review code
 - [ ] Check GitHub Actions build
+- [ ] If `icatdb Generator Script Consistency Test` CI job fails, is this because of a
+      deliberate change made to the script to change generated data (which isn't
+      actually a problem) or is there an underlying issue with the changes made?
 - [ ] Review changes to test coverage
 - [ ] {more steps here}
 
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -152,6 +152,7 @@ jobs:
 
   generator-script-testing:
     runs-on: ubuntu-16.04
+    continue-on-error: true
     name: icatdb Generator Script Consistency Test
     steps:
       - name: Add apt repo
@@ -262,6 +263,8 @@ jobs:
 
       # NOTE: If a delibrate change is made to the script that will change the data generated,
       # the diff (and therefore this job) will fail. If this is the case, don't be alarmed.
+      # The `continue-on-error` keyword has been added to this job so the workflow should
+      # pass when the PR is merged in, even if this job fails.
       # But, if you didn't mean to change the output of the script, there is likely a
       # problem with the changes made that will make the generator script's data
       # different to SciGateway preprod
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -248,6 +248,11 @@ jobs:
         with:
           ref: master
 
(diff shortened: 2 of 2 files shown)


sam-glendenning:
So what happens if you're making a deliberate change to the generator script?

MRichards99:
That's a decent point, something I hadn't really considered tbh. I guess we want to avoid making changes to the script that'll impact on the data generated (the aim is to keep preprod and other instances the same) but I guess if there's a case where a deliberate change needs to be made then we can just ignore that CI job?

Once the change gets merged into master, re-run the CI on master and it should pass (the test would diff output from master and master which will be identical). 

Perhaps not particularly clean when a deliberate change is required, but I feel it's important to monitor any changes of data between branches, unless there's another way of doing it that I haven't thought of?

sam-glendenning:
The CI either passes or fails and this seems like the sort of thing which would be ideal more as an alert that anything. Something like "alert: the data generated by this PR job differs from that of master". But since such a thing doesn't exist, it seems like the best thing to do is for any PR that you _expect_ a change in generated data to occur, make this clear in the PR description. Then you should get an expected 9/10 jobs passing.

louise-davies:
Maybe the techniques on this thread to allow that job to fail and also not affect the pull-mergability of a PR?: https://github.com/actions/toolkit/issues/399
Aka we can still see the job failed and can investigate it if it is unexpected, but if it is expected then it's easy to ignore

sam-glendenning:
I'd support that, using `continue-on-error` at least makes it clear that the step can fail without necessarily meaning the world is ending. However, I also heavily recommend making it clear in the PR description if you expect this to happen

MRichards99:
But since the diff is the final step of that job, would `continue-on-error` make any difference? Are we not able to merge PRs that have failing jobs, you just need an approving review right? I've added a couple comments/notes in 4c41b718e2ffd49c285025ddec21e0d86a84a9a5, perhaps this would help?

sam-glendenning:
I was mostly requesting for a note in the YAML specifying that this is a job that is "allowed" to fail in a sense (which continue-on-error can do) but it probably doesn't make much difference. Adding it in or not both result in a big ugly X. 

But you're right, you only need an approving review and as long as the reviewer has taken into account the changes being made, it should be fine.

sam-glendenning (approved):

sam-glendenning (approved):

PR #249, opened 2021-08-13
Ping Endpoint

### LINKED ISSUE #241: Implement Endpoint needed for Nagios Checks
**Description**:
When discussing Nagios checks with RIG, it was identified that we should have a 'the API responds' endpoint that doesn't require authentication. 

The endpoint should be `/ping` (to match IDS Server equivalent functionality and naming), with an output of `DataGateway API OK` if a 200 is returned.

For the Python ICAT backend implementation, I thought making use of `client.getEntityNames()` as this fetches entity names from ICAT and doesn't require the client object to be logged in. I should check that an unauthenticated client object can be fetched from the client cache-pool design used in the API.

For the database backend, doing a simple check should suffice. DB backend won't be put into production (so this implementation won't be used with Nagios) but we should implement it while we're there, just don't go down a rabbit hole with it, keep it simple

**Acceptance criteria**:
- [ ] A new endpoint should be implemented for both backends that allows for the API to be checked for responsiveness

MRichards99:
This PR will close #241 

## Description
This adds a `/ping` endpoint to both backends which can be used as a service check (for systems such as Nagios) when the API is in production. This endpoint doesn't require authentication (or any query params) so a request can be sent very easily. The endpoint does a basic check to ensure the communication method is alive and well. In both backends, it justs gets the table/entity names. This function call is surrounded by the library's most base exception (e.g. `ICATError`) - if an exception is caught, a 500 will be the response. 

A standard `DataGateway API OK` message will appear in the response body of both implementations (a message which has been put in `constants.py` to ensure if changed, all implementations change).

I've added docstrings, and within them, some OpenAPI/Swagger documentation where relevant. I've also generated `openapi.yaml` and pushed the additions.

There are a couple of basic tests for each implementation, just to check that the endpoint works and another that mocks a failed check and checks the relevant DataGateway API exception is raised (to cause a 500).

## Testing Instructions
Send a request to `/ping` and ensure you're happy with what you see

- [x] Review code
- [ ] If `icatdb Generator Script Consistency Test` CI job fails, is this because of a
      deliberate change made to the script to change generated data (which isn't
      actually a problem) or is there an underlying issue with the changes made?

## Agile Board Tracking
Connect to #241

### DIFF
--- a/datagateway_api/common/database/backend.py
+++ b/datagateway_api/common/database/backend.py
@@ -2,9 +2,14 @@ import datetime
 import logging
 import uuid
 
+from sqlalchemy import inspect
+from sqlalchemy.exc import SQLAlchemyError
+
 from datagateway_api.common.backend import Backend
+from datagateway_api.common.constants import Constants
 from datagateway_api.common.database.helpers import (
     create_rows_from_json,
+    db,
     delete_row_by_id,
     get_facility_cycles_for_instrument,
     get_facility_cycles_for_instrument_count,
@@ -20,7 +25,7 @@ from datagateway_api.common.database.helpers import (
     update_row_from_id,
 )
 from datagateway_api.common.database.models import SESSION
-from datagateway_api.common.exceptions import AuthenticationError
+from datagateway_api.common.exceptions import AuthenticationError, DatabaseError
 from datagateway_api.common.helpers import get_entity_object_from_name, queries_records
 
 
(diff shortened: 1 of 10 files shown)


datagateway_api/common/database/backend.py
```diff
@@ -32,6 +37,19 @@ class DatabaseBackend(Backend):
     Class that contains functions to access and modify data in an ICAT database directly
     """
 
+    def ping(self, **kwargs):
+        log.info("Pinging DB connection to ensure API is alive and well")
+
+        try:
+            inspector = inspect(db.engine)
+            print(type(inspector))
```
louise-davies, line 45: Is this meant to be here? Seems like a debug statement

MRichards99: Ah yes, I was working out the path of the function to mock. Removed in https://github.com/ral-facilities/datagateway-api/pull/249/commits/c9027505ec9ebd84923b5aa3bc6c96a82c1a6ae4

louise-davies (approved):

PR #273, opened 2021-10-18
Add Insensitive Like Operator to WHERE Filter

### LINKED ISSUE #243: Insensitive Like Operator
**Description**:
Just an issue for the action in https://github.com/ral-facilities/datagateway/issues/731. 

Once https://github.com/icatproject/python-icat/issues/86 has been solved, closed and merged, I need to update the `like` operator on DataGateway API to pass `UPPER()` into query. This will make our `like` operator case insensitive on production ICATs which use Oracle databases.

I already have these changes in a stash so I just need to write some tests to prove the where filter stores the upper function.

**Acceptance criteria**:
- [ ] Update like operator to use SQL upper function
- [ ] Add some tests to explicitly test this within the where filter

MRichards99:
This PR will close #243 

## Description
This adds two new operators for the WHERE filter, `ilike` and `nilike`. The first one will be used for https://github.com/ral-facilities/datagateway/issues/731, and the not like alternative was simple to implement while I was there so I might as well ready for the Search API. This makes use of JPQL functions, specifically `UPPER()` will is applied to both sides of the query, the values in icatdb as well as the value given by the user. The support for JPQL functions is being added in 0.20.0 of Python ICAT.

For https://github.com/ral-facilities/datagateway/issues/731 to be fixed, there will need to be a search and replace task on `like` to `ilike` in DataGateway. Equally, I'm happy to quickly change this PR so the `like` is actually case insensitive, but I believe having a specific `ilike` operator is the last thing we agreed on verbally. Something for @louise-davies and I to discuss.

I've had to edit some of the existing unit tests because there are some small changes in how conditions are stored in Python ICAT. The main change is that the conditions (stored in dictionary keys) are always stored as a list of strings, even if there is only a single condition per attribute. This makes no difference to functionality (nor am I concerned about this for anything in the future) but did require editing the tests to suit the new behaviour.

This PR will remain in a draft state until Python ICAT 0.20.0 has been released and I update the API's dependencies to use the new version.

I've also been trying to add a couple of tests that create a test DataGateway API (using Flask's test app) but pointing to ISIS dev ICAT instead of localhost. I've got it working independently of the rest of the tests, but it seems to get confused with the other test apps that run on localhost when running all tests together. I've got this saved in a stash and will try to get it working - don't want to spend excessive amounts of time on it.

## Testing Instructions
ICATs using mariadb are case insensitive by nature so this feature must be tested on an ICAT which uses an Oracle database. I have been using ISIS production and dev for my own manual testing, with the following requests:

```bash
# Should show two results
http://{{datagateway-api}}/investigations?where={"title": {"like": "tomato"}}

# Should also two results, even though the request sends toMATo
http://{{datagateway-api}}/investigations?where={"title": {"ilike": "toMATo"}}

# Should return two results as it's looking for things not like tomato
http://{{datagateway-api}}/investigations?where={"id": {"in": [12578343, 12578344]}}&where={"title": {"nlike": "toMATo"}}

# Should return an empty array as it matches the title with what's given in the request and NOTs that
http://{{datagateway-api}}/investigations?where={"id": {"in": [12578343, 12578344]}}&where={"title": {"nilike": "toMATo"}}
```

Until Python ICAT 0.20.0 is released, you'll need to build and install the fixes in Python ICAT manually.

- [x] Review code
- [x] If `icatdb Generator Script Consistency Test` CI job fails, is this because of a deliberate change made to the script to change generated data (which isn't actually a problem) or is there an underlying issue with the changes made?

## Agile Board Tracking
Connect to #243


MRichards99:
As this issue only affects production ICATs, I have just tested this by spinning up a local DataGateway by using ISIS prod config pointing at my local DataGateway API which also points at ISIS prod. You can see this works on the right (my local instance) whereas it doesn't work on the left (current deployment of DataGateway ISIS, without this fix).

image

MRichards99:
The safety part of the CI workflow will be fixed once #278 is merged.

VKTB (approved):
Looks good and manual tests pass as expected!

PR #274, opened 2021-10-25
Search API Structure

MRichards99:
## Description
Over the past week or so, I've doing some planning for the search API, understanding how things will be implemented into this repo. I've created a few commits to reflect this which show skeletons of files and unimplemented classes that will serve as placeholders so we know where to implement things when the time comes. 

Please see the bullet points in each of the commit messages that explain more about each change I've made.

@louise-davies I have assigned you as a reviewer. Please could you take a look (while the PR is in draft) at this branch/these commits and let me know your thoughts on the structure/skeleton bits I've added? I'm basically trying to split up DataGateway API and the Search API (see commit message of 568af6d) and then inherit/pick out functionality from the ICAT backend of DataGateway API to prevent duplication when creating the search API. This seems to be the least confusing way to structure this repository, but I might've missed something obvious so would appreciate your thoughts on this.

I do find the following path a bit jarring however:
`datagateway-api/datagateway_api/common/__init__.py`

Perhaps in an ideal world, we could rename `datagateway-api` to `icat-apis` (i.e. rename this repo) but I feel that might do more harm than good. Maybe best just to leave alone 🤔 

## Testing Instructions
Once we've discussed the structure, I'll go and make any changes, then fix the imports and turn this into a PR marked 'ready to review' and something that can be merged.

- [ ] Review code
- [ ] If `icatdb Generator Script Consistency Test` CI job fails, is this because of a
      deliberate change made to the script to change generated data (which isn't
      actually a problem) or is there an underlying issue with the changes made?


louise-davies (commented):
This all looks fairly sensible to me but I'm no expert in python project layouts. My only question is that there's a `common` folder and a `src` folder, both of which now have sections for each "backend"/"API". Is it then maybe a good idea to restructure so we don't need the separate `common` folder at the top level? Or maybe that's a bad idea. At the very least the naming confuses me - as `src` implies source code yet `common` also contains code. I think I understand it's meant to be `src` as the API layer and `common` as the business logic layer but I may be wrong.

Just my thoughts - feel free to ignore them as you & Viktor probably know better than I do due to you guys having more Python project experience than I

MRichards99:
I'm going to have a read around the how Python projects should be structured because I'm no expert either. But I think I agree with you, I find it confusing too. When there was much less code in `common/` (around the time I started work on the API) it made more sense, but there's now more code in `common/` than in `src/` which doesn't seem to make sense. Think it's one of those "it's always been like that" situations so I overlooked it.

MRichards99:
@louise-davies I've just pushed a few more commits on this so I ready for some further thoughts. I've essentially put `common/` into `src/` and removed some of the `datagateway_api/` and search_api/` because there was quite a few before.

Inside of `src/`, the following directories exist:
- `common/` - Files used between both APIs, such as config and the date handler
- `datagateway_api/` - Files specific to DataGateway API, all the backend and specific filter implementations
- `resources/` - Endpoint definition for both APIs
- `search_api/` - Files specific to the search API
- `swagger/` - Files for the swagger docs
- `main.py`, `config.json`, `wsgi.py` also live directly inside `src/`, there's no change from what currently happens on master

louise-davies (approved):
The new structure seems tidier and more understandable so I'm happy with it

Issue #283, opened 2021-11-03
Respect Config for Search API Endpoints


MRichards99:
**Description**:
While working on #257, both sets of endpoints (DataGateway API and search API) will always be created, the creation of the search API endpoints is [always carried out](https://github.com/ral-facilities/datagateway-api/blob/bc1d09d69aeb2378a7ec65a0f0350de3bb96deb5/datagateway_api/src/api_start_utils.py#L206). This should actually respect the config objects, which will be present once #256 is complete. This should just be editing the conditional to look up the relevant config object.

On #257, I also hardcoded the extension of search API endpoints to `/search_api`. Again, this should look at the config and use the extension defined in there. If #256 doesn't make use of the DataGateway API extension config option, this should also be added to those endpoints.

**Acceptance criteria**:
- [ ] DataGateway API and search API endpoints should only be created if the corresponding config objects are present in `config.json`. If neither are present, just create the endpoint for openAPI docs
- [ ] Use search API and DataGateway API extensions which are present in the config

Issue #256, opened 2021-09-23
Add DataGateway and PaNOSC Modes


MRichards99:
**Description**:
As DataGateway API and our implementation of the PaNOSC Search API will be developed in this repo, there needs to be a way of dealing with this. For the first release, I think it'll be best to add a configuration option for `datagateway` and `panosc` modes.

The `datagateway` mode will expose the current endpoints and the `panosc` mode will (upon implementing) expose the 11 endpoints for the search API. This config option will have to be made mandatory and can be done in `config.py`.

To do this, I need to add a config option (`mode`?) and use that option in `main.py` to branch off to the existing calls to `api_start_utils.py` and the new panosc function calls to setup the Flask app for that mode. To avoid `api_start_utils.py` from getting too long and confused, I might create a `search_api_start_utils.py` to have the endpoint definition and general app infrastructure setup there. At which point, it would probably make sense to rename the existing file to `datagateway_api_start_utils.py`.

In the future, we should look at supporting both at the same time (i.e. where a single instance can be run for both DataGateway API and PaNOSC Search API) by adding prefixes (perhaps something like `/datagateway/ENDPOINT` and `/panosc/ENDPOINT`) but that should come later.

**Acceptance criteria**:
- [ ] Add a way to switch between DataGateway and PaNOSC modes, which should be configurable
- [ ] Add `mode` config option to `config.json.example`
- [ ] Add test in to cover the changing of modes

MRichards99:
`[{"datagateway-api": "datagateway"}, {"panosc-api": "panosc"}]`

`[{datagateway-api": ""}]`

no prefix (blank string) is all good as long as length of array is 1

Check that no prefix strings match across the list

`{"datagateway": "datagateway-api", "panosc": "panosc-api"}`

Could flip them round to avoid collisions and have multiple instances of the same API? Could require further reworking of config if you wanted to point two DG API at two different ICAT instances.

PR #284, opened 2021-11-03
Define Endpoints for Search API

### LINKED ISSUE #257: Define Search API Endpoints
**Description**:
Each of the 11 endpoints can be defined in Flask without actually needing to implement them. The following endpoints should be added:

- GET /datasets/{pid}
- GET /datasets/{pid}/files
- GET /datasets/{pid}/files/count
- GET /datasets
- GET /datasets/count
- GET /documents/{pid}
- GET /documents
- GET /documents/count
- GET /instruments/{pid}
- GET /instruments
- GET /instruments/count

`test_endpoint_rules.py` tests the existing DataGateway API endpoints. There should be something similar for the Search API endpoints implemented with this issue.

**Acceptance criteria**:
- [ ] All endpoints are defined and can be pinged for a quick 200 (as they won't be implemented in this step)
- [ ] Add tests for endpoint rules, either extend the existing file or create a new one (depending on how I structure the files for the Search API across the repo)

MRichards99:
This PR will close #257

## Description
This PR defines endpoints which need to be implemented for the search API.

I've used the `endpoint=` argument for `add_resource()` to prevent naming collisions between endpoints. This was happening with endpoints like `/datasets` (DataGateway API endpoint) and `/search_api/datasets` even though they're actually different endpoints. Using `endpoint=` fixed this.

#256 isn't implemented yet so the [search API endpoints are always created](https://github.com/ral-facilities/datagateway-api/blob/bc1d09d69aeb2378a7ec65a0f0350de3bb96deb5/datagateway_api/src/api_start_utils.py#L206). Once the new config has been implemented, endpoint creation should respect the config and only create them when they are wanted by the user. I have also [hardcoded the extension](https://github.com/ral-facilities/datagateway-api/blob/bc1d09d69aeb2378a7ec65a0f0350de3bb96deb5/datagateway_api/src/api_start_utils.py#L208) which should be changed to look up the relevant config value once #256 is closed.

To remember to resolve both TODOs, I have opened #283 to tackle these issues once #256 and #257 have been completed.

## Testing Instructions
Run the API on this branch and ensure that you can send requests to each of the newly defined endpoints (e.g. `/search_api/datasets`) which return a 200 and a response body of `null`. 

I was a little undecided on whether this needs a version bump or not. I've gone for feature because these changes do add in new endpoints, even if they are unimplemented. I would welcome an alternative view on this though, I'm not sure if a different (or even no) version bump is more appropriate. 

- [x] Review code

## Agile Board Tracking
Connect to #257


VKTB (approved):
Looks good and all the endpoints return `null` as  expected. I feel as though a version bump is necessary though I am not sure which one should be used. Might be a good idea getting thoughts from others on this.

PR #285, opened 2021-11-04
Add DataGateway and PaNOSC Modes

### LINKED ISSUE #256: Add DataGateway and PaNOSC Modes

VKTB:
This PR will close #256 

## Description
Adds a configuration option for `datagateway` and `panosc` modes by separating the config so that DataGateway API and Search API are objects. With the existing logic, it was complicated to make these objects optional so that if neither are present only the openapi endpoint is created for example. Because of this, I decided to use the [pydanic](https://pydantic-docs.helpmanual.io/) library which allows for data validation and settings management using type annotations. With the help of this library, these two objects are optional by default but if they are present then all of their fields are mandatory with exception of the optional fields in the `datagateway_api` object. This is because some fields here may need to be mandatory depending on what backend is used. For example, if `backend` is set to `python_icat` then the icat related config values must be present, therefore with the help of the `require_icat_config_value` validator function, these fields become mandatory. The app exits if the validation fails - i.e. a mandatory field is missing, a field is misspelled or a field value is of wrong data type.

With these changes in place, we can modify the relevant logic so that only the endpoints for DataGateway API and Search API are only created if their config objects are present in the `config.json` file, otherwise only the endpoint for the Open API is created. Realised that issue https://github.com/ral-facilities/datagateway-api/issues/283 was created with respect to this, so the above can be addressed as part of it.

## Testing Instructions
Unit tests have been written to test the config though there is no harm if the manual tests at the bottom are performed.
Note: The test that was creating a temporary file was failing (even without the config changes) due to Windows file permission issue. A unit test should run in isolation and should not touch the file system at all therefore I instead mocked the `builtins.open` to fake `with open` so there is no more need for a temp file to be created and read.

- [x] Review code
- [x] The application should exit if:
  - [x] the `db_url` in the `datagateway_api` object is missing in `config.json` when `backend` is set to `db`
  - [x] any or all of the icat related fields are missing in the `datagateway_api` object when `backend` is set to `python_icat`
  - [x] the `config.json` file cannot be found
- [x] The application should not exit if:
  - [x] the `datagateway_api` and `search_api` objects are missing in `config.json`.
  - [x] the `db_url` in the `datagateway_api` object is missing in `config.json` when `backend` is set to `python_icat`
  - [x] any or all of the icat related fields are missing in the `datagateway_api` object when `backend` is set to `db`

## Agile Board Tracking
Connect to #256


MRichards99:
I will review this on Monday, but the first thing I notice is that one of the commit messages should have `BREAKING CHANGE:` because the old style of config won't work any more. 0cca696ac8e4cf42385d7a3dbd2deb0191cde52c or 2ffa84fc169969ebb260d10173125d83fc3f8754 might be the place for that.

datagateway_api/common/config.py
```diff
@@ -109,25 +88,85 @@ def set_backend_type(self, backend_type):
...
+
+    Some options used when running the API (host, debug_mode etc.) aren't mandatory
+    when running the API in production (these options aren't used in the `wsgi.py`
+    entrypoint). As a result, they're not present in `config_keys`. However, they
+    are required when using `main.py` as an entrypoint. In any case of these
+    specific missing config options when using that entrypoint, they are checked at
+    API startup so any missing options will be caught quickly.
+    """
+
+    datagateway_api: Optional[DatagatewayAPI]
```
MRichards99, line 142: Obviously this is marked by pydantic as optional (as it should be), but when I remove the `datagateway_api` object from `config.json`, the API errors on startup:

```bash
Traceback (most recent call last):
  File "/root/.pyenv/versions/3.6.8/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/root/.pyenv/versions/3.6.8/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/root/seg_dev/datagateway-api/datagateway_api/src/main.py", line 7, in <module>
    from datagateway_api.src.api_start_utils import (
  File "/root/seg_dev/datagateway-api/datagateway_api/src/api_start_utils.py", line 11, in <module>
    from datagateway_api.common.datagateway_api.backends import create_backend
  File "/root/seg_dev/datagateway-api/datagateway_api/common/datagateway_api/backends.py", line 4, in <module>
    from datagateway_api.common.datagateway_api.icat.backend import PythonICATBackend
  File "/root/seg_dev/datagateway-api/datagateway_api/common/datagateway_api/icat/backend.py", line 7, in <module>
    from datagateway_api.common.datagateway_api.icat.helpers import (
  File "/root/seg_dev/datagateway-api/datagateway_api/common/datagateway_api/icat/helpers.py", line 79, in <module>
    @cached(cache=ExtendedLRUCache())
  File "/root/seg_dev/datagateway-api/datagateway_api/common/datagateway_api/icat/lru_cache.py", line 22, in __init__
    super().__init__(maxsize=config.datagateway_api.client_cache_size)
AttributeError: 'NoneType' object has no attribute 'client_cache_size'
```

This is because of imports at the top of files, with `@cached(cache=ExtendedLRUCache())` making a call that looks at some DataGateway API specific config. Could you investigate if there's a way of refactoring the imports/code to stop this from happening please? The current behaviour means that a user cannot have config for just the search API. 

I wonder if this could be fixed by moving `get_cached_client()` into a class, maybe making that function into a static method? You might be able to think of a different/better solution so happy to see you fix it a different way.

datagateway_api/common/config.py
```diff
@@ -109,25 +88,85 @@ def set_backend_type(self, backend_type):
...
+    all required config options exist before getting too far into the setup of the API.
+
+    If a mandatory config option is missing or misspelled, or has a wrong value type,
+    Pydantic raises a validation error with a breakdown of what was wrong and the
+    application is exited.
+
+    Config options used for testing are not checked here as they should only be used
+    during tests, not in the typical running of the API.
+
+    Some options used when running the API (host, debug_mode etc.) aren't mandatory
```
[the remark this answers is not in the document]
MRichards99: Yes, as when the API is deployed for production use, they're set elsewhere (e.g. Apache is used to set the host and port). The following should be set to optional:
- `flask_reloader`
- `debug_mode`
- `host`
- `port`

I also wonder whether `test_user_credentials` and `test_mechanism` should be set as optional because these shouldn't be required for production use.

MRichards99 (changes requested):
The code looks good but there's a couple of things which would be good to resolve, including changing one of the commit messages to be a `BREAKING CHANGE:`

datagateway_api/common/config.py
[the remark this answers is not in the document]
VKTB: Thanks! Yes, `test_user_credentials` and `test_mechanism` should also be optional

datagateway_api/common/config.py
VKTB: Decorators run during import, before the function is even called. So by having `@cached(cache=ExtendedLRUCache())`, it means that an `ExtendedLRUCache` object will always be instantiated at the beginning even if that is inside a class. What I did instead is I set the `create_backend` function to be imported in `api_start_utils.py` only if the config contains a `datagateway_api` object.

VKTB:
> eview this on Monday, but the first thing I notice is that one of the commit messages should have `BREAKING CHANGE:` because the old style of config won't work any more. [0cca696](https://github.com/ral-facilities/datagateway-api/commit/0cca696ac8e4cf42385d7a3dbd2deb0191cde52c) or [2ffa84f](https://github.com/ral-facilities/datagateway-api/commit/2ffa84fc169969ebb260d10173125d83fc3f8754) might be the place for that.

The footer of be64faf9c2bb641d207a2b8a22a03ba68217682c now contains `BREAKING CHANGE:`.

datagateway_api/common/config.py
```diff
@@ -1,102 +1,81 @@
...
+    ValidationError,
+    validator,
+)
 
 
 log = logging.getLogger()
 
 
-class APIConfigOptions(Enum):
+class DatagatewayAPI(BaseModel):
```
MRichards99, line 20: I don't want to pick _that_ picky reviewer, but could you change this (and the two other instances of it in this file) to be `DataGatewayAPI` please?

MRichards99 (changes requested):
Thanks for making the changes, they fix the issues I noticed the other day. If you could change the `DatagatewayAPI` to `DataGatewayAPI` that would be great, and I'll approve it :)

MRichards99 (approved):
Looks good, if you could wait to merge until I've decided what to do on #274 so you can merge this straight into master, that would be great. Hopefully I'll get that sorted by the end of next week :)

MRichards99:
Because #274 has been merged into master, I'm happy for this to be merged. Fix the merge conflicts and check the CI passes then I'll give it the green tick so you can merge when you're ready :)

MRichards99 (approved):

PR #288, opened 2021-11-12
Add Study PIDs to Generator Script

### LINKED ISSUE #287: Add Study PIDs to Generator Script
**Description**:
As discussed in https://github.com/ral-facilities/datagateway/pull/909, it would be good to add some fake data for `study.pid` so DOI links can be tested. This change shouldn't impact the existing data generated as this means that someone has to go through the e2e tests on DataGateway and change them for the new data, it's not very fun!

**Acceptance criteria**:
- [ ] Ensure `study.pid` is generated, using the same way as DOIs
- [ ] The change doesn't change the existing data

MRichards99:
This PR will close #287 

## Description
These changes add PIDs to studies, in the same format as the script deals with `investigation.doi`.

Using the same instance of Faker that's used in the rest of the script would impact all the other data which is something we want to avoid. As a result, I've create a new instance of Faker and fed the input seed to that. I _think_ that's the best way to deal with this now and in the future?

Upon merging, the new version will be v1.0.1.

## Testing Instructions
Look for differences between the study data on SciGateway preprod and the data generated from this script. I have done this and the only differences are the additions of new DOIs, no other changes. I sent the following the request to preprod and my local instance and diffed them using a very nice online tool:
`http://{{datagateway-api}}/studies?order="id asc"&limit=10`

image

The generator script job on the CI will fail but the only differences should be on the study data (only the PIDs).

- [x] Review code

## Agile Board Tracking
Connect to #287

### DIFF
--- a/datagateway_api/common/database/models.py
+++ b/datagateway_api/common/database/models.py
@@ -1229,6 +1229,7 @@ class STUDY(Base, EntityHelper, metaclass=EntityMeta):
     modId = Column("MOD_ID", String(255), nullable=False)
     modTime = Column("MOD_TIME", DateTime, nullable=False)
     name = Column("NAME", String(255), nullable=False)
+    pid = Column("PID", String(255))
     startDate = Column("STARTDATE", DateTime)
     status = Column("STATUS", Integer)
     userID = Column("USER_ID", ForeignKey("USER_.ID"), index=True)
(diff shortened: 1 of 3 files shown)


MRichards99:
Just noticed an addition to the swagger docs so pushed that change. I have no need to make any further changes before a review :)

VKTB (approved):

PR #289, opened 2021-11-15
Add PaNOSC Search API Data Model

### LINKED ISSUE #264: Add PaNOSC Search API Data Model
**Description**:
Similar to DataGateway API's database backend, the data model should be defined using classes etc. You can see the DB backend modelling of ICAT in `datagateway_api/common/database/models.py`. This should be as simple as using the [data model](https://github.com/panosc-eu/search-api/blob/master/doc/data-model.md) defined in the example implementation. We have discussed using [pydantic](https://pydantic-docs.helpmanual.io/) for this which looks like it could be useful.

Within these classes, the relationships between entities need to be defined. In the database backend, SQLAlchemy did this nicely using `Column()`. Obviously we can't use that library in this case, so we should consider how this can be defined. Maybe just defining the fields as listed in the docs will be enough so long as you can validate what values they are set as - I guess Pydantic will help with that.

There should probably be tests to cover each class/entity of the data model, to ensure they are implemented correctly. I guess it needs to be tested that each attribute of an entity can be fetched/set using sensible values?

**Acceptance criteria**:
- [ ] Implement data model like database backend `models.py`
- [ ] Add tests for each class of the data model

VKTB:
This PR will close #264 

## Description
This PR makes changes to the PaNOSC data models so that they use pydantic which allows for data validation and settings management using python type annotations. It enforces type hints at runtime, and provides user friendly errors when data is invalid.

There is no need to test the the class models as they are defined with help of pydantic therefore this is already tested by pydantic as that is its responsibility to do. The class models themselves are basically schemas so they do not deal with creation but only define what data the models expect. There will be a separate where these models will be referenced and their creation will occur, and this will be the place where the creation will be tested.

## Testing Instructions

- [ ] Review code
- [ ] Implement data model like database backend `models.py`

## Agile Board Tracking
Connect to #264


MRichards99:
Based on our discussions last week, I don't think we'll need to convert from PaNOSC to ICAT. As such, I think you can remove the `to_icat()` functions from each of the classes.

We do need to think about where to convert the inputs of filters however. The contents of a WHERE filter will contain entity names for the PaNOSC data model and we will need to convert those inputs into something that ICAT can understand. Maybe around the time of `apply_filter()` is an appropriate place to do this; this is not the domain of this issue/PR but just to update you with my thoughts from what I was working on last week.

MRichards99:
I've just pushed the changes - there is an F407 failure from flake8 (`future feature annotations is not defined`). This is because this wasn't introduced to the standard library of Python until Python 3.7. It doesn't look like the import is used though?

VKTB:
> I've just pushed the changes - there is an F407 failure from flake8 (`future feature annotations is not defined`). This is because this wasn't introduced to the standard library of Python until Python 3.7. It doesn't look like the import is used though?
> 
> Edit: It hasn't appeared on GitHub Actions because that's running 3.9 on the linting job but it was present on my `nox -s lint` which runs 3.6.

It is not explicitly used but it makes `.update_forward_refs()` work. It errors without this import.

MRichards99:
It seems that import is not backwards compatible, there's no alternative import or external library that will backport it for 3.6. Would [self-referencing models](https://pydantic-docs.helpmanual.io/usage/postponed_annotations/#self-referencing-models) be an alternative solution that'll work on all Python 3 versions?

VKTB:
> It seems that import is not backwards compatible, there's no alternative import or external library that will backport it for 3.6. Would [self-referencing models](https://pydantic-docs.helpmanual.io/usage/postponed_annotations/#self-referencing-models) be an alternative solution that'll work on all Python 3 versions?

As per the documentation you linked, it looks like not-yet constructed models can be referenced using a string in 3.6. I quickly texted it on an example implementation locally and it seemed to work. So essentially what we will have to do is remove the annotation import and go back to the `Affiliation`, `Dataset`, `Document` and `Member` models (as they are the only models that reference not-yet constructed models) and reference the models using a string. For example, in the `Affiliation` model, we will need to change `members: Optional[List[Member]]` to `members: Optional[List["Member"]]`.

MRichards99:
Yes that looks like the best way around it, if you could make the changes that would be great.

VKTB:
As requested, I addressed the changes discussed and also made changes to the models to match the updated data-model guide.

MRichards99 (approved):

PR #290, opened 2021-11-16
Respect Config for Search API Endpoints

### LINKED ISSUE #283: Respect Config for Search API Endpoints

VKTB:
This PR will close #283

## Description
The DataGateway API and Search API endpoints are now only created if the corresponding config objects (`datagateway_api` and `search_api`) are present in `config.json`. If neither are present, just the endpoint for openAPI docs is created. The end parts of the endpoint URLs are now contain the API extensions provided in the config.

https://github.com/ral-facilities/datagateway-api/pull/290/commits/81fa46907673d95e954be275acdb7b28b02892ae is a `BREAKING CHANGE` commit as API clients will be affected by this change.

## Testing Instructions

- [x] Review code
- [x] DataGateway API and search API endpoints should only be created if the corresponding config objects are present in `config.json`. If neither are present, just create the endpoint for openAPI docs
- [x] Search API and DataGateway API endpoint extensions are used

## Agile Board Tracking
Connect to #283


VKTB:
Louise asked if any of the APIs can be served at `/` when a blank extension is provided in the config. With the changes at the moment this will cause issues because the API extension is injected between two slashes in the endpoint URL - `f"/{datagateway_api_extension}/sessions",`. This should not happen so I will quickly remove the slash before the extension variable in the endpoint URLs but this would mean that the extension itself has to be prefixed with a slash in the config file. This can be enforced using pydantic in PR https://github.com/ral-facilities/datagateway-api/pull/285.

MRichards99:
@VKTB your proposed solution sounds sensible, requiring the slash prefix in the configuration shouldn't be a problem if you can get pydantic to help with this.

datagateway_api/src/api_start_utils.py
```diff
@@ -181,16 +208,66 @@ def create_api_endpoints(flask_app, api, spec):
...
 
         # Ping endpoint
         ping_resource = ping_endpoint(backend, client_pool=icat_client_pool)
-        api.add_resource(ping_resource, "/ping")
+        api.add_resource(ping_resource, f"{datagateway_api_extension}/ping")
         spec.path(resource=ping_resource, api=api)
 
+    # Search API endpoints
+    if config.search_api is not None:
+        # TODO - Use config value when new config style is implemented
```
MRichards99, line 224: This TODO can be removed because you've used the config value :)

test/datagateway_api/icat/filters/test_order_filter.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 from typing_extensions import OrderedDict
 
+
```
MRichards99, line 4: Flake8 isn't complaining about this but the standard gap between import groups is a single line. Could this additional line be removed please? I'm guessing this managed to sneak its way in by accident :)

MRichards99 (changes requested):
This work looks good, and I can see that one of the APIs can be served on `/` if this is configured as such. There's 2 tiny style changes I've suggested and I think your branches need merging (on this branch and `add-datagateway-and-panosc-modes-#256`) to get the CI passing but I'd be happy to approve once this is done.

MRichards99 (dismissed):

MRichards99 (approved):
Ready to merge when you're ready

VKTB:
> Ready to merge when you're ready

@MRichards99 This can now be merged.

PR #291, opened 2021-11-22
BREAKING CHANGE: Merge pull request #285 from ral-facilities/add-data…

MRichards99:
PR to fix the lack of version bump when merging #285


MRichards99:
I don't think this will work, I'm looking into a different fix for this

PR #292, opened 2021-11-22
Query Parameters Inputs for Search API

### LINKED ISSUE #259: Get Query Param (Filters for Querying)
**Description**:
The Search API uses different syntaxes for filters than is implemented in DataGateway API. For example, the Search API puts all the filters into a single query param whereas DataGateway API splits them up per filter type. To deal with these differences, an
alternative implementation of the `QueryFilterFactory` and `get_filters_from_query_string()` should be created.

For the alternate version of `get_filters_from_query_string()`, the majority of endpoints accept `filter` as the query param key, so that's what the function should look for. However, the count endpoint look for `where` as the query param key, so this
difference should be implemented here.

The format/syntax of each filter is here: https://github.com/panosc-eu/search-api/blob/master/doc/query.md

**Acceptance criteria**:
- [ ] Add code to get filters from request into the code (`get_filters_from_query_string()`) supporting `filter` and `where` query params
- [ ] Use the request data to create `PythonICATXXXFilter` objects (`QueryFilterFactory`) - this is a good foundation for the proceeding work to implement each one of the filters even if it's decided that we need new filter classes which inherit from the ICAT backend filter classes
- [ ] Write tests that cover each format listed in `query.md`

MRichards99:
This PR will close #259 

## Description
This is a draft PR for the work on the query parameters. I've done most of the work on this and have written all the test cases required for this work, so the failures on `TestSearchAPIQueryFilterFactory` dictate the remaining work for this issue. 

In summary, the following needs to be completed:
- Text operator on where filter - I've implemented the text operator on the include filter but I didn't realise it would actually work directly on the where filter. The factory class currently has no indication of which entity a filter comes from, which is required for text operator. This can be done on the include filter because you have the related entity to work with. Something like `entity_name` needs to be passed into the static method to make this happen
- Nested and/or - see failing tests for this.

ee668e38cd43354851163616a93924ad84e14b90 contains `feat` commit type so the correct version bump should be made.

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Connect to #259


datagateway_api/src/search_api/query_filter_factory.py
```diff
@@ -0,0 +1,151 @@
...
+                                "where": {
+                                    full_field_path: related_model["scope"]["where"][
+                                        where_key
+                                    ],
+                                },
+                            },
+                        }
+                        query_filters.extend(
+                            SearchAPIQueryFilterFactory.get_query_filter(where_filter,),
+                        )
```
VKTB, line 128: @MRichards99 Correct me if I am wrong but my understanding here is that two `SearchAPIWhereFilter` objects will be created, one after another, if there are two elements in `field_names`. However, wouldn't this mean that both of these filter objects will need to find a match for the provided value? 

For example, imagine we were querying the `/Datasets` endpoint with a relation `instrument` and _WHERE_ filter `"text": "ESS" `. The text operator fields for `instrument` are `name` and `facility`, meaning that there will be one WHERE filter for `instrument.name` and another for `instrument.facility`. Wouldn't this mean that it will look for `instrument.name` with value `ESS` **AND** `instrument.facility` with value `ESS`?

Shouldn't it be instead looking for `instrument.name` with value `ESS` **OR** `instrument.facility` with value `ESS`? If so then I am guessing we can create an _OR WHERE_ filter here and send that to `get_query_filter`?

datagateway_api/src/search_api/filters.py
```diff
@@ -10,8 +10,9 @@
 
 
 class SearchAPIWhereFilter(PythonICATWhereFilter):
-    def __init__(self, field, value, operation):
+    def __init__(self, field, value, operation, boolean_operator="and"):
```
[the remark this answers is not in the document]
MRichards99: Yes I think we should move code where needed but I think there's a balance. If we move `PythonICATWhereFilter` to `common/`, do we move `DatabaseWhereFilter` too? If we move one but not the other, it means that `datagateway_api/src/datagateway_api/icat/` & `datagateway_api/src/datagateway_api/database/` won't be even in that one will have `filters.py` but the other won't. `PythonICATOrderFilter` won't be used by the search API so does that mean we keep that where it is or move it to `common/`? 

I think there are cases where we should move DG API code to `common/` but I don't think the filters should be moved - I think that might add to the confusion.

datagateway_api/src/search_api/query_filter_factory.py
```diff
@@ -0,0 +1,151 @@
...
+                        # TODO - we might want to move this to the data
+                        # definitions at a later point
+                        text_operator_fields = {
+                            "datasets": ["title"],
+                            "documents": ["title", "summary"],
+                            "files": ["name"],
+                            "instrument": ["name", "facility"],
+                            "samples": ["name", "description"],
+                            "techniques": ["name"],
+                        }
```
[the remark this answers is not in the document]
MRichards99: I agree, there will be work needed to make that dictionary work for singular and plural field names.

As for my comment regarding data definitions, my point was that the dictionary probably shouldn't be 'hardcoded' in that factory class. It should be put elsewhere - be that in the same file where the PaNOSC data model is defined (that's what I meant by "data definitions"), or in `constants.py`.

datagateway_api/src/search_api/query_filter_factory.py
MRichards99: I agree, the text operator should create where filters using the OR boolean operator. Good spot

datagateway_api/src/search_api/filters.py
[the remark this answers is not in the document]
VKTB: Valid point, I did not think about the DB class filters..

datagateway_api/src/search_api/query_filter_factory.py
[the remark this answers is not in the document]
VKTB: I agree that they should not be hardcoded here. It might be best to leave it as TODO for now because there might be something nicer that we can do as part of the data models themselves which may not require singular and plural dict. I am struggling to picture how that may work at this point but it may become a bit clearer later in the project. We will just have to remember that it works for some cases as it is but not for all and revisit it at some point.

datagateway_api/src/search_api/query_filter_factory.py
```diff
@@ -0,0 +1,159 @@
...
+    SearchAPISkipFilter,
+    SearchAPIWhereFilter,
+)
+
+log = logging.getLogger()
+
+
+class SearchAPIQueryFilterFactory(QueryFilterFactory):
+    @staticmethod
+    def get_query_filter(request_filter, entity_name=None):
```
VKTB, line 17: We need to think about how we get `entity_name` for an endpoint. This is easy for when the Text operator is part of an _INCLUDE_ filter because the entity_name is essentially the value specified in the `relation` field. However, what about when there is only a _WHERE_ filter that has a Text operator?

datagateway_api/src/search_api/query_filter_factory.py
```diff
@@ -0,0 +1,159 @@
...
+                where_filters.extend(
+                    SearchAPIQueryFilterFactory.get_query_filter(
+                        where_filter, entity_name,
+                    ),
+                )
+            except KeyError:
+                # Do not raise FilterError nor attempt to create filters. Simply
+                # ignore text operator queries on fields that are not part of the
+                # text_operator_fields dict.
+                pass
```
VKTB, line 107: This is what the example Search API does.

test/search_api/test_search_api_query_filter_factory.py
```diff
@@ -0,0 +1,477 @@
...
+                    },
+                },
+                "parameters",
+                2,
+                ["name", "value"],
+                ["like", "gte"],
+                ["Test name", 275],
+                ["and", "and"],
+                id="Nested AND, multiple conditions with operator",
+            ),
```
VKTB, line 140: I am really confused how a similar OR AND nested query will work.. There is an example one [here](use OR boolean operator for WHERE filters in text operator #259).

With the logic as it is, all WHERE filter objects will end up having an `OR` boolean operator but I don't think this is what we want.

We want this:

`(WHERE Filter 1 AND WHERE Filter 2) OR (WHERE Filter 3 AND WHERE Filter 4)` and not <br>`WHERE Filter 1 OR WHERE Filter 2 OR WHERE Filter 3 OR WHERE Filter 4`.

Text operators inside `AND` will also be affected by this.

VKTB:
@MRichards99 It would be nice to get your thoughts on the changes I made and some of the things I flagged as part of the review comments.

datagateway_api/src/search_api/query_filter_factory.py
MRichards99: I've addressed this for you in 1d24021ce32fe7fa9843de7c438aa6713d4eb44e. You can now use `entity_name` passed into `get_query_filter()` to know which entity name is used.

datagateway_api/src/search_api/query_filter_factory.py
MRichards99: I'm slightly confused, are you agreeing with the comment or do we need to do something with this?

datagateway_api/src/search_api/query_filter_factory.py
```diff
@@ -94,40 +124,17 @@ def get_include_filter(filter_input):
...
-                                "where": {
-                                    full_field_path: related_model["scope"]["where"][
-                                        where_key
-                                    ],
-                                },
-                            },
-                        }
-                        query_filters.extend(
-                            SearchAPIQueryFilterFactory.get_query_filter(where_filter,),
+                # Scope filter can have WHERE, INCLUDE, LIMIT and SKIP filters
```
MRichards99, line 128: Are there examples of there being include, limit and skip filters inside the scope property? Or do they work in the example implementation? I assumed only a WHERE is expected inside the scope, but maybe all the filters work?

datagateway_api/src/search_api/query_filter_factory.py
VKTB: Thanks for that!

datagateway_api/src/search_api/query_filter_factory.py
VKTB: This was meant to serve as a note in case you were wondering why we are not doing anything here. It is what the example Search API does so I have done the same. We don't need to do anything.

test/search_api/test_search_api_query_filter_factory.py
MRichards99: Why would all where filters default to having an OR boolean operator? [They default to `and`](https://github.com/ral-facilities/datagateway-api/blob/1d24021ce32fe7fa9843de7c438aa6713d4eb44e/datagateway_api/src/search_api/filters.py#L13), they'll only be set to OR if that's passed in at the creation of a `SearchAPIWhereFilter`. Unless you're referring to the nested stuff, maybe we should talk about that over Zoom so we walk through what's going on.

As an aside, I would say the test case which you've attached this comment to is a bit misleading/isn't really nested (my fault) because the top level AND only has one item in it (which contains and AND of 2 items). So you've got `(A AND B) AND` - there's nothing on the right hand side (RHS) of the top-level AND. Do you want me to add a couple of test cases to make this clearer?

I'd probably add the following test cases for nested boolean operators:
- `(A AND B) AND (C AND D)` - in boolean algebra, this is basically just `A AND B AND C AND D`
- `(A AND B) OR (C AND D)`
- `(A OR B) OR (C OR D)` - in boolean algebra, this is basically just `A OR B OR C OR D`
- `(A OR B) AND (C OR D)`
- `(A AND B AND C) OR (C OR D) OR (E AND F)`

Feel free to create the test cases but I'm happy to write them if makes your work easier, just let me know.

datagateway_api/src/search_api/query_filter_factory.py
VKTB: Not in the _Querying_ mark down but I did find a few examples of the include filter being inside the scope property in the [Document queries](https://github.com/panosc-eu/search-api/blob/master/doc/document-example-queries.md) mark down. This prompted me into looking at the [Loopback 4 Node.js](https://loopback.io/doc/en/lb4/Include-filter.html#include-with-filters) documentation where I discovered that scope supports limit and skip filters in addition to where and include. While looking for the LoopBack link just now, I also noticed that fields is another filter listed in there. I can confirm that all of these filters work in the example Search API implementation.

MRichards99: Good spot, I hadn't realised this. Does your change [here](https://github.com/ral-facilities/datagateway-api/blob/cdb30ff0d4ebc8a0901e385d14bbd623f7c96fbb/datagateway_api/src/search_api/query_filter_factory.py#L128) mean we support all filters in the `scope` property now? Sadly this means adding tests that they work... I know `test_valid_include_filter()` is getting quite long now (well all the pytest params) so it might be worth splitting that up between 'regular' include filters (i.e. just includes) and then another test function for all the scope stuff. I'll leave the decision to you, just an option if that function get a bit unweldy to maintain :)

test/search_api/test_search_api_query_filter_factory.py
VKTB: I thought I updated my comment which I initially wrote in rush but looks like I forgot. The example hyperlink does not also link to the OR AND nested query that I intended to link you to, so apologies for the confusion.

The reason I attached my comment to this test case is because there is an example nested OR AND query that is similar to this case that can be found [here](https://github.com/panosc-eu/search-api/blob/master/doc/dataset-example-queries.md#filter-2). If you look at it, we want to end up having `(A AND B) OR (C AND D)`. With the current implementation, the WHERE filter objects will take the closest logical operator to them and use that as a boolean operator. In the example case, that will be the `and` logical operator but this would mean that we will end up having `A AND B AND C AND D`. With the way the filters are implemented at the moment, I am confused about how to get from that to `(A AND B) OR (C AND D)`. 

Regarding this:
>With the logic as it is, all WHERE filter objects will end up having an `OR` boolean operator...

I was referring to the filter objects in the linked example case and not all. The reason this happens is because the example case is a nested OR AND query so the changes I added [here](https://github.com/ral-facilities/datagateway-api/blob/1d24021ce32fe7fa9843de7c438aa6713d4eb44e/datagateway_api/src/search_api/query_filter_factory.py#L74-L75) modify the boolean operators of the WHERE filters objects to be equal to the logic operators that is the farthest from them which is the `or` logical operator in this case. Now. this is not what we want but there is another reason why I made these changes and I only found about the consequences that it will have with nested logical operator queries after I tried to write a nested OR AND unit test. Essentially the reason was so that we do not have to repeat what we are already doing in the `else` part of the conditional. Instead of creating WHERE filter objects in the `if` part of the conditional, we just create a WHERE filter query and pass that to `get_query_filter` so that the `else` part can deal with the creation of WHERE filter objects. See https://github.com/ral-facilities/datagateway-api/pull/292/commits/a5905148e524fec250c1d381440cca0dc3fab7a7 for more. This works fine for non-nested but does not for nested logical operator queries.

I am happy creating the test cases that you suggested but we need to have a chat on Zoom about the outlined issue.

datagateway_api/src/search_api/query_filter_factory.py
VKTB: >Does your change [here](https://github.com/ral-facilities/datagateway-api/blob/cdb30ff0d4ebc8a0901e385d14bbd623f7c96fbb/datagateway_api/src/search_api/query_filter_factory.py#L128) mean we support all filters in the `scope` property now?

Correct. Essentially, it takes whatever is inside `scope`, creates a new filter request and passes it to `get_query_filter` to deal with it because `get_query_filter` already knows how to deal with WHERE, INCLUDE, LIMIT and SKIP filters.

test/search_api/test_search_api_query_filter_factory.py
```diff
@@ -0,0 +1,2098 @@
...
+                    [],
+                    ["instrument.name", "eq", "Instrument 1", "and"],
+                    [],
+                    [],
+                    [],
+                ],
+                [None, None, None, None, None, None, None, None, 2, 50, None],
+                [None, None, None, None, None, None, None, None, None, None, 20],
+                id="Complex case",
+            ),
```
VKTB, line 2033: In this test case, we end up with two `SearchAPILimitFilter` objects but how do we know which one is for the `datasets` entity and which one is for the `instrument` relation? I do not think we can work this out. Maybe the `SearchAPILimitFilter` class should have another field called `entity_name` or something else that will indicate to which entity the LIMIT filter is related to?

The same applies to SKIP filters.

VKTB:
I wrote quite a lot of test cases as part of https://github.com/ral-facilities/datagateway-api/pull/292/commits/8f90ac5049ab9e7f009c894029d9a3bc1626b929, https://github.com/ral-facilities/datagateway-api/pull/292/commits/0f0691782c10a2f805cf11d8ca8d90d0db4e6b36 and https://github.com/ral-facilities/datagateway-api/pull/292/commits/0f0691782c10a2f805cf11d8ca8d90d0db4e6b36. There are a lot of edge cases with the query parameters inputs which is why there are so many cases. We may discover down the line that some of them are redundant but at this point I think that having some redundant tests is probably better than not enough tests.

VKTB (commented):
The tests that are failing are nested logical operators and nested relation tests. The former will be fixed once the logic that deals with nested logical operators is implemented. It will likely require making changes to the tests to get them to pass with the nested logical operator changes that will be introduced. The latter tests require some thinking - please see my comment below for explanation and suggestion.

VKTB:
@MRichards99 I noticed something with the text operator while I was going through the deliverable PDF. If you look at the [dataset query example](https://github.com/panosc-eu/search-api/blob/master/doc/dataset-example-queries.md#query-datasets-with-files-matching-a-string-using-full-text-search) (which looks like is the only query example that uses a text operator), you will see that it queries for `"text": "file1"`. If you compare this with the response below it, you will see that the `name` value of the file returned is `panosc-file1.hdf` and not `file1` which indicates that the `like` instead of the `eq` operator is used in text operators. With the changes in this PR, we create an `SearchAPIWhereFilter` instance that has an `eq` operator when there is a text operator in the query and this is what our tests expect the `SearchAPIWhereFilter` objects to have. 

Unless there is something that I am not aware of then we should change this. I can make the change while I am fixing failing nested filter tests.

MRichards99:
> @MRichards99 I noticed something with the text operator while I was going through the deliverable PDF. If you look at the [dataset query example](https://github.com/panosc-eu/search-api/blob/master/doc/dataset-example-queries.md#query-datasets-with-files-matching-a-string-using-full-text-search) (which looks like is the only query example that uses a text operator), you will see that it queries for `"text": "file1"`. If you compare this with the response below it, you will see that the `name` value of the file returned is `panosc-file1.hdf` and not `file1` which indicates that the `like` instead of the `eq` operator is used in text operators. With the changes in this PR, we create an `SearchAPIWhereFilter` instance that has an `eq` operator when there is a text operator in the query and this is what our tests expect the `SearchAPIWhereFilter` objects to have.
> 
> Unless there is something that I am not aware of then we should change this. I can make the change while I am fixing the failing nested filter tests.

Yes that's true, if you could change that, that would be great!

test/search_api/test_search_api_query_filter_factory.py
VKTB: We decided not to support LIMIT and SKIP filters on relations so marking this as resolved.

VKTB:
I fixed the tests that were failing in the `TestSearchAPIQueryFilterFactory` class. I also made changes so that:
- only one `SearchAPIIncludeFilter` object is created per nested include filter
- the `like` operation is used in `SearchAPIWhereFilter` objects when they are created from text operators
- `FilterError` is raised if the scope filter contains skip and limit filters.
- the `get_condition_values` function always returned a list with one tuple inside it therefore it made sense to modify it so that it returns a tuple.

MRichards99:
I'm happy with these changes but I don't think we can merge them because it relies on changes in Python ICAT which haven't made it to an official release. I've tried to change that dependency to a specific git commit (rather than a version number) but it Poetry didn't seem to want to do it.

VKTB (approved):

MRichards99:
@VKTB f61ed00c69853e4b4d00d5d610c93dcd5aa9057a makes a small change to use the new implementation of the `where_clause` property as added in https://github.com/icatproject/python-icat/pull/100. This is really to make the tests happy on this branch (they'll pass once we can specify the new Python ICAT release), with the bulk of the changes being made in #298

VKTB (approved):

PR #293, opened 2021-11-24
Extension Validation Bugfix

MRichards99:
## Description
This PR fixes a bug I found with `validate_extension()` when trying one of the APIs on `/`. I've also added some unit tests for this function so this can be avoided in the future.

## Testing Instructions

- [ ] Review code

### DIFF
--- a/datagateway_api/src/common/config.py
+++ b/datagateway_api/src/common/config.py
@@ -27,10 +27,11 @@ def validate_extension(extension):
     """
     extension = extension.strip()
 
-    if not extension.startswith("/"):
-        raise ValueError("must start with '/'")
-    if extension.endswith("/"):
-        raise ValueError("must not end with '/'")
+    if extension:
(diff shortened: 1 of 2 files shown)


VKTB (changes requested):
I like your approach of testing the function directly, I should have opted for this when I wrote the tests. We can remove the `test_load_with_invalid_api_extension_does_not_start_with_slash` and `test_load_with_invalid_api_extension_ends_with_slash` tests because the tests you added cover these cases. In addition to this, I also noticed one tiny thing which I flagged in my comment below.

Looks good otherwise.

MRichards99:
I think we should keep the tests that you reference because they test at a different angle i.e. with actual test config data, looking to see if SystemExit occurs in an invalid case. I've committed your suggested change so it actually fails at the point I wanted it too, not before

VKTB (approved):

PR #294, opened 2021-11-26
Search API Session/Client Handling

### LINKED ISSUE #258: Solve Session Handling for Search API
**Description**:
The Search API will not have any authentication from the client side. However, Python ICAT requires credentials/valid session ID to access the majority of its functionality. Using the anon authenticator will be the solution for this. However, we need a mechanism for creating a valid session and then passing clients round the code.

The Search API should login to create a session ID at start-up (using `client.login()`). This session ID should either be used for the duration of the API's uptime (we would have to use `client.refresh()` to ensure the session doesn't expire) or create a new
session ID every two hours.

We might be able to get away with using a single client across the Search API (as I tried for testing purposes when figuring out DataGateway API's client handling) but this should be load tested, for collisions and for speed. If it is not suitable, then making
use of the object pool specifically would be a good alternative. I don't believe the cache element will be needed because the session ID will be the same for each user so I'm not sure if it's worth having that part of the design.

**Acceptance criteria**:
- [ ] Add the ability for a session ID to be created and maintained throughout the API's uptime
- [ ] Investigate which client handling strategy would be best for the Search API
- [ ] Integrate the best client handling method into the Search API (so they can be used in the requests, like DataGateway API)

MRichards99:
This PR will close #258 

## Description
This PR adds a solution for client handling for the search API. Compared with DataGateway API, this is a much more basic solution; because the search API will only ever use the anon user in ICAT, I think a single client object can be used for the API. This client is stored in a `SessionHandler` class which is accessed by the endpoints when they need it. 

The session for this client is managed by the `client_manager` decorator. This is a similar design to `requires_session_id` on DataGateway - i.e. a decorator to be used for endpoints. The client manager decorator looks at the remaining lifetime of the session and if the session has expired, the client will log back in as an anon user. There's no initial login/session ID being set as this use case (i.e. the first time `SessionHandler.client`) is used, a `ICATSessionError` will be caught due to the empty `sessionId` attribute in the client.

I've made use of `ICATClient` for the search API to reduce duplication - this class inherits Python ICAT's `Client` but extends it slightly for the cache object pool design. `client_use` now allows for client configuration for both search API and DataGateway API. 

In terms of testing, I've done some mocking of config (really to test that `client_use` does what I expect it to) so I've moved a couple of pytest fixtures out to `conftest.py`. I've also had to put the config into its own class in `config.py` because otherwise, it cannot be mocked. This is because when `config.py` is imported, the old way (i.e. no in a class) is set as soon as its imported which makes it impossible to mock.

The commits tell a 'nice story' (they're not all over the place and confusing like usual!) so I won't go through each change one-by-one, hopefully looking through each commit gives you a good understanding.

This is a new 'feature', so the new version of the search API will be 3.1.0.

## Testing Instructions
I've done some testing using Apache JMeter and it passes, with no drop in speed (all requests were <600ms on my machine) - 200 'users' are spun up and each send a request to `/datasets`. I'm happy to give the file to the reviewer if this helps.

It should be noted that the ICAT anon authenticator is required for all the tests to pass. This is because `SessionHandler.client` will login using that authenticator. I've made sure the CI has this authenticator by [making a change in ICAT Ansible](https://github.com/icatproject-contrib/icat-ansible/pull/71).
- [x] Review code

## Agile Board Tracking
Connect to #258

### DIFF
--- a/test/datagateway_api/icat/test_icat_client.py
+++ b/test/datagateway_api/icat/test_icat_client.py
@@ -1,10 +1,8 @@
-import json
-from unittest.mock import mock_open, patch
+from unittest.mock import patch
 
 from icat.client import Client
 import pytest
 
-from datagateway_api.src.common.config import APIConfig
 from datagateway_api.src.datagateway_api.icat.icat_client_pool import ICATClient
 
 
--- a/test/datagateway_api/icat/test_icat_client.py
+++ b/test/datagateway_api/icat/test_icat_client.py
@@ -1,5 +1,10 @@
-from icat.client import Client
+import json
+from unittest.mock import mock_open, patch
 
+from icat.client import Client
+import pytest
+
+from datagateway_api.src.common.config import APIConfig
 from datagateway_api.src.datagateway_api.icat.icat_client_pool import ICATClient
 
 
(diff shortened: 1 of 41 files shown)


MRichards99:
I've fixed the issues on the CI which is now passing.

test/datagateway_api/icat/test_icat_client.py
```diff
@@ -10,6 +15,47 @@ def test_init(self):
...
+            with patch(
+                "datagateway_api.src.common.config.Config.config", api_config,
+            ):
+                with patch(
+                    "icat.client.Client.__init__", side_effect=MockClient.__init__,
+                ):
+                    test_icat_client = ICATClient(client_use)
+                    assert test_icat_client.url == expected_url
+                    assert test_icat_client.checkCert == expected_check_cert
+
```
VKTB, line 58: Is it not possible to make use of `test_config` here? Is the print statement needed in the `__init__`?

```suggestion
    def test_client_use(
        self, test_config, client_use, expected_url, expected_check_cert,
    ):
        class MockClient:
            def __init__(url, checkCert=True):  # noqa
                print(f"URL: {url}, Cert: {checkCert}")
                # Would've preferred to assign these values to self but this didn't
                # seem to be possible
                Client.url = f"{url}/ICATService/ICAT?wsdl"
                Client.checkCert = checkCert

        with patch(
            "datagateway_api.src.common.config.Config.config", test_config,
        ):
            with patch(
                "icat.client.Client.__init__", side_effect=MockClient.__init__,
            ):
                test_icat_client = ICATClient(client_use)
                assert test_icat_client.url == expected_url
                assert test_icat_client.checkCert == expected_check_cert

```

test/datagateway_api/icat/test_icat_client.py
```diff
@@ -10,6 +13,44 @@ def test_init(self):
...
+                id="Search API Usage",
+            ),
+        ],
+    )
+    def test_client_use(
+        self, test_config, client_use, expected_url, expected_check_cert,
+    ):
+        class MockClient:
+            def __init__(url, checkCert=True):  # noqa
+                print(f"URL: {url}, Cert: {checkCert}")
```
VKTB, line 38: Just repeating my second question from my previous review comment. Is the `print` statement here needed?
```suggestion
```

MRichards99: Oh I thought the original commit suggestion deleted it, thanks for rechecking

VKTB (approved):

PR #295, opened 2021-12-02
Implement Limit and Skip filters

### LINKED ISSUE #262: Implement Search API Limit Filter
**Description**:
The limit filter follows the same syntax as DataGateway API, though is part of the all-in-one query parameter. This shouldn't require too much work, just integrating it with the Search API.

**Acceptance criteria**:
- [ ] Get limit filter working with Search API

### LINKED ISSUE #263: Implement Search API Skip Filter
**Description**:
The skip filter follows the same syntax as DataGateway API, though is part of the all-in-one query parameter. This shouldn't require too much work, just integrating it with the Search API.

**Acceptance criteria**:
- [ ] Get skip filter working with Search API

VKTB:
This PR will close #262 
This PR will close #263 

## Description
The Limit (`SearchAPILimitFilter`) and Skip (`SearchAPISkipFilter`) filters are already implemented. To reduce duplication, they inherit from `PythonICATLimitFilter` and `PythonICATSkipFilter`. The `PythonICATSkipFilter` class connects to ICAT to get the `maxEntities` value applying the Skip filter to the ICAT query so this PR refactors the `PythonICATSkipFilter` class to take a `client_use` value in order to allow for client configuration for both Search API and DataGateway API.

It also adds tests which test the `SearchAPILimitFilter` and `SearchAPISkipFilter` classes.

## Testing Instructions
- [ ] Review code

## Agile Board Tracking
Connect to #262
Connect to #263

### DIFF
--- a/datagateway_api/src/datagateway_api/icat/filters.py
+++ b/datagateway_api/src/datagateway_api/icat/filters.py
@@ -210,13 +210,20 @@ class PythonICATOrderFilter(OrderFilter):
 
 
 class PythonICATSkipFilter(SkipFilter):
-    def __init__(self, skip_value):
+    def __init__(self, skip_value, client_use="datagateway_api"):
         super().__init__(skip_value)
+        self.client_use = client_use
 
     def apply_filter(self, query):
-        icat_properties = get_icat_properties(
-            config.datagateway_api.icat_url, config.datagateway_api.icat_check_cert,
-        )
+        if self.client_use == "datagateway_api":
+            icat_properties = get_icat_properties(
+                config.datagateway_api.icat_url, config.datagateway_api.icat_check_cert,
+            )
+        else:
+            icat_properties = get_icat_properties(
+                config.search_api.icat_url, config.search_api.icat_check_cert,
+            )
+
         icat_set_limit(query, self.skip_value, icat_properties["maxEntities"])
 
 
--- a/datagateway_api/src/datagateway_api/icat/filters.py
+++ b/datagateway_api/src/datagateway_api/icat/filters.py
@@ -210,12 +210,12 @@ class PythonICATOrderFilter(OrderFilter):
 
 
 class PythonICATSkipFilter(SkipFilter):
-    def __init__(self, skip_value, client_use="datagateway_api"):
+    def __init__(self, skip_value, filter_use="datagateway_api"):
         super().__init__(skip_value)
-        self.client_use = client_use
+        self.filter_use = filter_use
 
     def apply_filter(self, query):
-        if self.client_use == "datagateway_api":
+        if self.filter_use == "datagateway_api":
             icat_properties = get_icat_properties(
                 config.datagateway_api.icat_url, config.datagateway_api.icat_check_cert,
             )
--- a/datagateway_api/src/search_api/filters.py
+++ b/datagateway_api/src/search_api/filters.py
@@ -19,7 +19,7 @@ class SearchAPIWhereFilter(PythonICATWhereFilter):
 
 class SearchAPISkipFilter(PythonICATSkipFilter):
     def __init__(self, skip_value):
-        super().__init__(skip_value)
+        super().__init__(skip_value, client_use="search_api")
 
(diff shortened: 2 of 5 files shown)


VKTB:
The tests inside the `TestSearchAPILimitFilter` and `TestSearchAPISkipFilter` classes are based off the test that are already in `TestICATLimitFilter` and `TestICATSkipFilter`. However, I noticed that the `test_valid_limit_value` and `test_valid_skip_value` tests are integration rather than unit tests. This is because they connect to ICAT and as a result take ~705ms and ~751ms respectively to complete. We can keep these as integration tests and separate them together with other integration tests for when we plan to do this in future. But if we want to unit test the `SearchAPILimitFilter` and `SearchAPISkipFilter` classes, this is how we should do it (they take ~3ms to complete):

`TestSearchAPILimitFilter`:
```Python
@pytest.mark.parametrize(
    "limit_value",
    [
        pytest.param(10, id="typical"),
        pytest.param(0, id="low boundary"),
        pytest.param(9999, id="high boundary"),
    ],
)
@patch.object(Query, "setLimit")
def test_valid_limit_value(self, query_mock, limit_value):
    test_filter = SearchAPILimitFilter(limit_value)
    test_filter.apply_filter(query_mock)

    query_mock.setLimit.assert_called_once_with((0, limit_value))
```

`TestSearchAPISkipFilter`:
```Python
@pytest.mark.parametrize(
    "skip_value", [pytest.param(10, id="typical"), pytest.param(0, id="boundary")],
)
@patch("datagateway_api.src.datagateway_api.icat.filters.get_icat_properties", return_value={"maxEntities":5000})
@patch.object(Query, "setLimit")
def test_valid_skip_value(self, query_mock, get_icat_properties_mock, skip_value):
    test_filter = PythonICATSkipFilter(skip_value)
    test_filter.apply_filter(query_mock)
    limit_value = get_icat_properties_mock()["maxEntities"]

    query_mock.setLimit.assert_called_once_with((skip_value, limit_value))
```

datagateway_api/src/datagateway_api/icat/filters.py
```diff
@@ -210,13 +210,20 @@ def apply_filter(self, query):
 
 
 class PythonICATSkipFilter(SkipFilter):
-    def __init__(self, skip_value):
+    def __init__(self, skip_value, client_use="datagateway_api"):
```
MRichards99, line 213: Perhaps another name for `client_use` might be more appropriate? `filter_use` or an alternative you can think of

datagateway_api/src/search_api/filters.py
```diff
@@ -19,7 +19,7 @@ def apply_filter(self, query):
 
 class SearchAPISkipFilter(PythonICATSkipFilter):
     def __init__(self, skip_value):
-        super().__init__(skip_value)
+        super().__init__(skip_value, client_use="search_api")
```
MRichards99, line 22: Same comment regarding `client_use` as above.

MRichards99 (changes requested):
Implementation looks good, I think it might be benefical to change `client_use` to a name more focused on filters.

As for the tests, I think let's keep them as they are now (i.e. lets focus on the search API deadline) and refactor/group the tests when there is time. I think I blur the line between unit and integration tests quite a lot (mainly for speed to write the tests) but this might not be the optimal way to test. I would like this repo to have a set of unit tests which can be run without a working ICAT instance attached to it but I think the higher priority is delivering a solution for the search API. I think the mixture of tests that are in this repo do give good coverage though, and are ultimately better than the small amount of testing that used to be present in this repo. TL;DR - the testing serves its purpose to know we're not breaking stuff, but could do with a refactor to make adhere to 'proper' developer practices.

VKTB:
Sounds good about the tests :)

MRichards99 (approved):

MRichards99:
Approved because I know you can fix merge conflicts :)

MRichards99 (approved):
I forgot that adding a commit will dismiss my review, so maybe second time lucky

PR #301, opened 2022-01-10
Implement Include Filter for Search API

### LINKED ISSUE #261: Implement Search API Include Filter
**Description**:
Like the WHERE filter, the include filter uses a different syntax to what we have in DataGateway API. Syntax: `{"include": [{"relation": "relatedModel"}]}`. This should really be covered when dealing with the query param changes but there might be
additional changes.

The new piece of functionality for this filter is the `scope` property. This allows for 'match conditions' to be made, which essentially just allows to add a WHERE filter into the include filter. I can't find any examples of other filters used in this `scope`
property, but we should maybe clarify that/test the example implementation. This shouldn't be too hard to do, upon finding `scope`, create a WHERE filter and add it to the query.

As always, test cases should be written for this filter to cover the basic usage and the scope aspect.

**Acceptance criteria**:
- [ ] Ensure basic syntax works
- [ ] Add functionality for `scope` property
- [ ] Write test cases to cover the above

MRichards99:
This PR will close #261 

## Description
This PR implements the include filter for the search API. The logic to this is fairly simple and similar to the where filter actually - convert the PaNOSC field names to ICAT field names and then let Python ICAT add the relevant stuff to the JPQL query (via the code in `PythonICATIncludeFilter`).

I've moved `get_icat_mapping()` into `PaNOSCMappings` as it seems to be more logically placed there. In doing so, I removed the logic that deals with the special use case where the ICAT mapping is a list of field names and put it in `apply_filter()`. I've done this to keep `get_icat_mapping()` a bit more generic, though we can move it back at a later stage if needed (if we find that the logic needs to be executed in multiple places). 

I have also refactored the query params code. The include filter can support a list of related fields, but the query params only created include filters for a single related entity so multiple `SearchAPIIncludeFilter`s would be created when only one is necessary. I've made the query params able to create include filters with multiple related fields (as proved by the diff of `test_search_api_query_filter_factory.py` in eab7f3d9d10eb36a998653a492c99476fc7f46d9).

## Testing Instructions
I guess similar to #298, the method of manual testing would be to send some requests to `/search-api/datasets` with include filters and see what you can break. If the filter is successfully created, the response body will show the version of ICAT server which is just a placeholder until the endpoint is actually implemented. As above, not all the mappings will work because some need ICAT 5.0 (currently unreleased) but I've added them in so we don't forget about them.

There is a feature version bump. I'm unsure of the exact numeric version bump there'll be because we have a few PRs awaiting to be merged but `poetry run semantic-release print-version` shows that the version bump does work (currently bumped to 3.2.0 as the latest version is 3.1.1). 

- [ ] Review code

## Agile Board Tracking
Connect to #261


VKTB (changes requested):

datagateway_api/src/search_api/panosc_mappings.py
```diff
@@ -26,6 +26,49 @@ def __init__(
...
+        :raises FilterError: If a valid mapping cannot be found
+        """
+
+        log.info(
+            "Searching mapping file to find ICAT translation for %s",
+            f"{panosc_entity_name}.{field_name}",
+        )
+
+        try:
+            icat_mapping = mappings.mappings[panosc_entity_name][field_name]
```
VKTB, line 52: This should be like the below otherwise the method needs to be static
```suggestion
            icat_mapping = self.mappings[panosc_entity_name][field_name]
```

datagateway_api/src/search_api/query_filter_factory.py
```diff
@@ -238,17 +245,23 @@ def get_include_filter(include_filter_input, entity_name):
...
+                        for included_entity_pointer in range(
+                            len(scope_query_filter.included_filters),
+                        ):
+                            nested_include = True
+                            included_filter = scope_query_filter.included_filters[
+                                included_entity_pointer
+                            ]
+                            scope_query_filter.included_filters[
+                                included_entity_pointer
+                            ] = f"{included_entity}.{included_filter}"
```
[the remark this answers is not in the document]
MRichards99: Good spot, I've also added and committed the addition of the assertion so we have a test that monitors this for the future.

test/search_api/filters/test_search_api_include_filter.py
```diff
@@ -0,0 +1,109 @@
...
+            # TODO - Below test fails because dataset/investigation parameters issue on
+            # mapping
+            pytest.param(
+                SearchAPIIncludeFilter(["dataset.parameters.document"], "File"),
+                "File",
+                "SELECT o FROM Datafile o INCLUDE o.dataset AS d, d.parameters AS p,"
+                " p.investigation",
+                id="File.dataset.parameters.document",
+                marks=pytest.mark.skip,
+            ),
```
VKTB, line 65: Like we discussed, a `DatasetParameter` entity does not have an `investigation` field so this test can be removed.
```suggestion
```

VKTB (commented):
The recent changes that you made look fine. It's just the test that you marked as skip that I think can be removed, see my comment. Also, the tests I listed in https://github.com/ral-facilities/datagateway-api/pull/298#pullrequestreview-863865894 are failing but that is not an issue as it can be sorted in that PR.

test/search_api/filters/test_search_api_include_filter.py
MRichards99: Agreed, just removed it

PR #303, opened 2022-01-19
Implement Search API WHERE Filter Operators

### LINKED ISSUE #297: Implement Search API WHERE Filter Operators
**Description**:
Following on from #260, some operators need to be added to the WHERE filter. These should be added to `PythonICATWhereFilter` because we might as well add support for DataGateway API.

The following needs to be added:
- `between` - @VKTB investigated that it is supported in Python ICAT, just needs adding into the API and tests written
- `inq` - we already have `in`, just need a conditional to change `self.operation` to `in` then call `create_filter()` again. This should be done so we don't cause a breaking change for DataGateway
- `nin` - @VKTB has investigated this, just needs adding with tests written
- `neq` - we already have `ne`, so add this operator in the same way I've described for `inq`
- `regexp` @VKTB investigated this, just needs adding in with tests written

**Acceptance criteria**:
- [ ] Implement `between`, `inq`, `nin`, `neq`, `regexp` as operators without causing a breaking change

VKTB:
This PR will close #297 

## Description
Implements `between`, `inq`, `nin`, `neq` and `regexp` operators in the `PythonICATWhereFilter` class without causing a breaking change.

## Testing Instructions
- [ ] Review code
- [ ] Test that the new operators are working on any of the DG API endpoints.

## Agile Board Tracking
Connect to #297


datagateway_api/src/common/filters.py
```diff
@@ -27,11 +27,23 @@ def __init__(self, field, value, operation):
...
+                    f"When using the {self.operation} operation for a WHERE filter, the"
+                    f" values must be in a list format e.g. [1, 2, 3]",
+                )
+
+        if self.operation == "between":
+            if not isinstance(self.value, list):
+                raise BadRequestError(
+                    "When using the 'between' operation for a WHERE filter, the values"
+                    " must be in a list format e.g. [1, 2]",
+                )
```
MRichards99, line 42: This looks like duplicated code. Could you add `"between"` to the list on L30 and remove this duplicated code here?

datagateway_api/src/common/filters.py
```diff
@@ -27,11 +27,23 @@ def __init__(self, field, value, operation):
...
         self.operation = operation
 
-        if self.operation == "in":
+        if self.operation in ["in", "nin", "inq"]:
             if not isinstance(self.value, list):
                 raise BadRequestError(
-                    "When using the 'in' operation for a WHERE filter, the values must"
-                    " be in a list format e.g. [1, 2, 3]",
+                    f"When using the {self.operation} operation for a WHERE filter, the"
+                    f" values must be in a list format e.g. [1, 2, 3]",
```
MRichards99, line 34: If making the change suggested below, you might want to change the example list to `[1, 2]` to match that a between should only have two values given to it.

datagateway_api/src/datagateway_api/icat/filters.py
```diff
@@ -51,6 +51,9 @@ def create_filter(self):
             where_filter = self.create_condition(self.field, "=", self.value)
         elif self.operation == "ne":
             where_filter = self.create_condition(self.field, "!=", self.value)
+        elif self.operation == "neq":
+            self.operation = "ne"
+            where_filter = self.create_filter()
```
MRichards99, line 56: I know this is how I suggested to implement, but your code in `src/common/filters.py` has made me think of a different, perhaps better way. So checking a list of operations, rather than just one, changing `self.operation` and calling the same function again:

```suggestion
        elif self.operation in ["ne", "neq"]:
            where_filter = self.create_condition(self.field, "!=", self.value)
```

datagateway_api/src/datagateway_api/icat/filters.py
```diff
@@ -88,6 +91,28 @@ def create_filter(self):
                 self.value = "(NULL)"
 
             where_filter = self.create_condition(self.field, "in", self.value)
+        elif self.operation == "inq":
+            self.operation = "in"
+            where_filter = self.create_filter()
```
MRichards99, line 96: If we decide the above suggestion is the better way to do it, please make the same change to this operator too.

MRichards99 (changes requested):
Looks good, have a think about the suggestion I made, do you think its the better way? It didn't come to me until I saw your other code, I think it should work? If you think its better then make the change, otherwise we can stay with the initial implementation if there's a reason why we shouldn't make the change.

datagateway_api/src/common/filters.py
VKTB: That will work, thanks.

VKTB:
> Looks good, have a think about the suggestion I made, do you think its the better way? It didn't come to me until I saw your other code, I think it should work? If you think its better then make the change, otherwise we can stay with the initial implementation if there's a reason why we shouldn't make the change.

Initially, I had the code like you suggested but after reading your description of the issue, I thought there was a reason behind adding a conditional to change `self.operation` and then `create_filter()` again. As a result of this, I modified the code to reflect this and did not test whether doing it the initial way breaks anything.

I just applied your suggestions and it seems to work fine. I manually tested the operators and the tests in `/test/datagateway_api/icat` pass too.

MRichards99 (approved):
Looks good! Like usual, let's leave this PR unmerged until we get the Python ICAT stuff sorted, then I can go through our mountain of PRs and merge them in one go.

Issue #312, opened 2022-01-31
Make Number of Years to Determine Public Data Configurable


MRichards99:
**Description**:
In `models.py`, the `isPublic` field is determined by a date check, currently hardcoded to 3 years as this is the time for ISIS. We should have a config option to set the number of years that determines public data so this can be changed for other facilities or if the ISIS rules change.

**Acceptance criteria**:
- [ ] Implement a config option for years for `isPublic` field
- [ ] Add config tests if the existing ones don't already cover this

PR #323, opened 2022-02-04
Cannot Use Filters on isPublic Field

### LINKED ISSUE #308: Cannot Use Filters on isPublic Field
**Description**:
Consider the following filter:
`filter={"where": {"isPublic": true}}`

When querying an endpoint using this filter, a 400 is returned. This is because the mapping for `isPublic` isn't direct, we use the field to know where to calculate from. I think there's a bit of confusion about this, either in the query param stage or when a filter with `isPublic` is applied to the query.

I think the calculation has been put in place in the form of `set_is_public()` so it could just be resolving some bug in the query param logic.

One difficulty of `isPublic` would be on count endpoints:
`where={"isPublic": true}`

When we query ICAT using `COUNT`, we will only ever get a number back. From that, I don't we can calculate which ones are public. In this case, it might be better to assume all results returned are public; the search API is only dealing with public data (hence there's no authentication and we use the anon user to login). The API should return a 200 and the count from ICAT, instead of giving a 400.

**Acceptance criteria**:
- [ ] Get `isPublic` to work as a field on a where filter, for both non-count and count endpoints
- [ ] Unskip tests relating to this issue in `test_count_endpoint.py` and `test_search_endpoint.py`

### LINKED ISSUE #322: Where filter without operator part does not work with integer and boolean values
**Description**:
Sending the following request:
`http://{{datagateway-api}}/search-api/datasets/0-8401-1070-7?filter={"include": [{"relation": "files", "scope": {"where": {"size": 1}}}]}`

returns `400 BAD REQUEST`:
```
local variable 'value' referenced before assignment
```

This errors only when integer or boolean values are used, but works fine with string values.

**Acceptance criteria**:
- [ ] Where filter without operator part should work with integer and boolean values

VKTB:
This PR will close #308, closes #322

## Testing Instructions
- [ ] Review code
- [ ] {more steps here}


datagateway_api/src/search_api/query_filter_factory.py
```diff
@@ -327,3 +343,34 @@ def prefix_where_filter_field_with_entity_name(where_filters, entity_name):
...
+        them will have a value that is boolean too. As a result of this, the value and
+        operation need to be converted to a format appropriate for the mapped ICAT
+        field. In the example above, if the filter asks for all Datasets whose isPublic
+        fields are set to True to be returned, then this method changes the WHERE
+        filter value to a three years ago datetime value and the operator to less than
+        so that all Datasets older than 3 years (which ISIS considers public) are
+        returned.
+        """
+        current_datetime = datetime.now(timezone.utc)
+        three_years_ago = current_datetime - relativedelta(years=3)
```
VKTB, line 365: This will be made configurable in #312

MRichards99 (approved):

PR #331, opened 2022-02-14
Search API Error Signalling/Formatting

### LINKED ISSUE #296: Search API Error Signalling
**Description**:
In the search API deliverable document, it shows the format that should be followed for the response body when an error occurs:

```json
{
    "error": {
        "statusCode": 400,
        "name": "Error",
        "message": "Cannot parse JSON-encoded object value."
    }
}
```

Explanations of the following keys (extracted from the deliverable document):
 - "statusCode": An HTTP status code of either 4xx or 5xx
 - "name": The name of the error
 - "message": A descriptive error message, to aid the user in troubleshooting

I think implementing this similarly to how `@queries_records` works is the simplest option to apply this for just the search API and across each of the endpoints.

**Acceptance criteria**:
- [ ] When a 4xx or 5xx code is given in response to a request, follow the format above

MRichards99:
This PR will close #296 

## Description
This PR adds a decorator so 4xx/5xx responses are formatted as per the specification (see the linked issue for what that should be). This decorator is applied to each of the search API endpoint functions in `search_api_endpoints.py`. This is a similar solution to handling errors for DataGateway API (`@queries_records`). 

## Testing Instructions
The automated tests follow the same mechanism as the `queries_records` decorator so I've the same thing for this code. For manual testing, try and get 4xx/5xx responses and check the response body matches the format in the original issue.

`poetry run semantic-release print-version` shows 4.1.0 which is what I expect the version bump to be.

- [ ] Review code

## Agile Board Tracking
Connect to #296

### DIFF
--- a/datagateway_api/src/api_start_utils.py
+++ b/datagateway_api/src/api_start_utils.py
@@ -3,6 +3,7 @@ import logging
 from pathlib import Path
 
 from apispec import APISpec
+from flask import Response
 from flask_cors import CORS
 from flask_restful import Api
 from flask_swagger_ui import get_swaggerui_blueprint
@@ -55,7 +56,12 @@ class CustomErrorHandledApi(Api):
     """
 
     def handle_error(self, e):
-        return str(e), e.status_code
+        if isinstance(e.args[0], (str, dict, tuple, Response)):
+            error_msg = e.args[0]
+        else:
+            error_msg = str(e)
+
+        return error_msg, e.status_code
 
 
 def create_app_infrastructure(flask_app):
--- a/datagateway_api/src/resources/search_api_endpoints.py
+++ b/datagateway_api/src/resources/search_api_endpoints.py
@@ -9,6 +9,7 @@ from datagateway_api.src.search_api.helpers import (
     get_files_count,
     get_search,
     get_with_pid,
+    search_api_error_handling,
 )
 
 log = logging.getLogger()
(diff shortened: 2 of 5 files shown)


VKTB:
Consider the following request:
`http://{{datagateway-api}}/search-api/documents?`

it returns:
```json
{
    "error": {
        "message": "attempt to return more than 10000 entities",
        "name": "PythonICATError",
        "statusCode": 500
    }
}
```
While the error message that comes from ICAT is useful (as it is) to someone using the Search API, I don't think the error name is. They probably do not know anything about ICAT so the error name is not relevant to them. I think we just need to catch `PythonICATError` in our calls to ICAT and raise an exceptions that is more suitable (perhaps `SearchAPIError`)? I do not expect finding the places where this error is raised to be an easy job because Python sadly does not enforce the handling of exceptions like Java does with checked exceptions.
 
The following request:
`http://{{datagateway-api}}/search-api/datasets?filter={"include": [{"relation": "datasets"}]}`

returns:
```json
{
    "error": {
        "message": "Bad PaNOSC to ICAT mapping: ('datasets',)",
        "name": "FilterError",
        "statusCode": 400
    }
}
```
I think the error message needs to be reworded because, again, someone using the Search API probably does not know anything about ICAT.

MRichards99:
I can see your point, but I'm not sure where the `name` of the response body will actually be used in the federated search frontend. I'm reluctant to spend time on neatening up the names unless there's demand for it from the frontend - I feel like there's tasks of higher priority to be working on. 

Also, if we are sent bug reports when the API is used in production, I guess they'll send us the response body so having the actual API exception might actually be more useful for us? It might make it easier to investigate and resolve said bugs

VKTB:
>I'm reluctant to spend time on neatening up the names unless there's demand for it from the frontend - I feel like there's tasks of higher priority to be working on.
I totally agree which is why I said it should be addressed separately at some point in future (if needed) rather than as part of this PR.

PR #340, opened 2022-02-22
Search API Datetime Format

### LINKED ISSUE #338: Search API Example Date Format not Working
**Description**:
When pointing the search API to ISIS dev ICAT (or using the deployed instance), the following request works as expected:
`http://{{datagateway-api}}/search-api/datasets?filter={"limit": 10, "where": {"creationDate": "2007-09-12 14:05:23+01:00"}}`

That uses the date format used by DataGateway API. However, if you see an [example search API query](https://github.com/panosc-eu/search-api/blob/master/doc/dataset-example-queries.md#response), the following date format is used in the response:
`2020-05-05T15:01:02.341Z`

If you use that format in the request, it doesn't work. See the following request as an example:
`http://{{datagateway-api}}/search-api/datasets?filter={"limit": 10, "where": {"creationDate": "2007-09-12T14:05:23.000Z"}}`

That request gives an internal ICAT exception. Some more details:
```
Exception Description: The object [2007-09-12T14:05:23.000Z], of class [class java.lang.String], from mapping
[org.eclipse.persistence.mappings.DirectToFieldMapping[createTime-->DATASET.CREATE_TIME]] with descriptor
[RelationalDescriptor(org.icatproject.core.entity.Dataset --> [DatabaseTable(DATASET)])], could not be converted to
[class java.sql.Timestamp].
```

To me, that implies the datetime is not in a format which ICAT can understand so we need to use the format that DataGateway API uses instead (the one in the first request) upon querying ICAT.

I will contact Carlo and see how important this is, but thought I'd document what I found.

**Acceptance criteria**:
- [ ] TBU upon contacting Carlo

MRichards99:
This PR will close #338 

## Description
This PR allows multiple datetime formats to be used on WHERE filters on the search API and aligns the output of datetime data to that of SciCat's implementation of the search API. This change only impacts the search API - I didn't want to change anything for DataGateway API in case it broke something for the frontend.

The email chain regarding datetime has most of the details about this

Version bump will be to 4.2.0. 

## Testing Instructions
Just try to break datetimes I guess, be aware that the original DataGateway API format doesn't work on the search API but it doesn't look like the federated search will use that.

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Connect to #338


test/search_api/test_models.py
```diff
@@ -12,123 +11,123 @@
 }
 
 DATASET_ICAT_DATA = {
-    "endDate": "2000-12-31 00:00:00+00:00",
+    "endDate": "2000-12-31T00:00:00.000Z",
```
VKTB, line 14: Does Python ICAT returns `datetime` values in this format? The tests in this module convert data from ICAT to PaNOSC format so the `datetime` values in the ICAT data dictionaries should be in the correct format. Not sure if the old format is the right one as I cannot remember whether I looked this up when I wrote the tests so best to double check.

MRichards99: The original format is the Python ICAT format, I changed it because the `DateHandler` conversion functions (str to datetime I think?) didn't seem to work. So in an attempt to just get something to work, I just changed all the datetime string values to one that the conversion actually can convert. To move back to the original, `DateHandler` will have to be changed so it can convert that format. That would be the 'best practice' solution, but I'm not sure how time consuming it'll be to change it.

VKTB: I see why you changed it. If you use `DateHandler.str_to_datetime_object`, which I initially used, then that only converts a `string` to a `datetime` object whereas the updated tests now expect the actual `datetime` values to be of type `string`. To get it to work I think that we just need to convert the `datetime` object generated by `DateHandler.str_to_datetime_object` to a `string` but in a Search API format. Doing this would allow us to change the `datetime` values back to the original format.

If we take the `creationDate` key in the `DATASET_PANOSC_DATA` dictionary as an example, I think that we can get get it to work like this:
```python
"creationDate": DateHandler.str_to_datetime_object(DATASET_ICAT_DATA["createTime"]).strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3] + "Z"
```

or by importing `SearchAPIDatetime` from `models.py` and doing this if we want to keep the `datetime` `string` format in one place:
```python
"creationDate": SearchAPIDatetime.use_search_api_format(DateHandler.str_to_datetime_object(DATASET_ICAT_DATA["createTime"]))
```

MRichards99: I removed https://github.com/ral-facilities/datagateway-api/commit/6242fe1e9ede39ed696feb567c4238e29dde6eb6 and made the change using `use_search_api_format()`. On data which is in the PaNOSC data model (e.g. the changes in `test_search_endpoint.py`), I've searched and replaced those

VKTB: Thanks for sorting this @MRichards99, glad it worked. Looks good, approving now.

VKTB (approved):

Issue #319, opened 2022-02-03
Test Search API on ISIS ICAT


MRichards99:
**Description**:
Once the endpoints have been finished and merged, we should probably point the search API at ISIS ICAT dev and do a bit of manual testing. Send a request for each type of endpoint (for each entity too) to check they all work. Send requests with some where and include filters to ensure conditions work and related entities return as expected.

**Acceptance criteria**:
- [ ] Test search API when pointed to ISIS ICAT

MRichards99:
#314 will fix various issues I've noticed on /datasets and /instruments (due to ISIS having no PIDs on these entities)

MRichards99:
@VKTB this looks to be caused by the differences between the ICAT schema and the PaNOSC data model. By using limit and skip to target certain records, that request does work. The following request (the same as yours but has a limit of 1 and skip of 2) does work:

`http://{{datagateway-api}}/search-api/documents?filter={"limit": 10}&filter={ "include": [ { "relation": "members", "scope": { "where": { "role": { "neq": "principal_experimenter" } }, "include": [ { "relation": "person" } ] } }, { "relation": "datasets" } ], "limit": 1, "skip": 2}`

However, if you use a skip of 3, you get an error. Looking at my logs (I'm using my local instance pointing at ISIS dev ICAT), there's a `ValidationError`:

```bash
Traceback (most recent call last):
  File "/root/seg_dev/datagateway-api/datagateway_api/src/search_api/helpers.py", line 38, in wrapper_error_handling
    return method(*args, **kwargs)
  File "/root/seg_dev/datagateway-api/datagateway_api/src/resources/search_api_endpoints.py", line 34, in get
    return get_search(entity_name, filters), 200
  File "/root/seg_dev/datagateway-api/datagateway_api/src/search_api/session_handler.py", line 38, in wrapper_client_manager
    return method(*args, **kwargs)
  File "/root/seg_dev/datagateway-api/datagateway_api/src/search_api/helpers.py", line 104, in get_search
    panosc_record = panosc_model.from_icat(icat_data, entity_relations).json(
  File "/root/seg_dev/datagateway-api/datagateway_api/src/search_api/models.py", line 244, in from_icat
    return super(Document, cls).from_icat(icat_data, required_related_fields)
  File "/root/seg_dev/datagateway-api/datagateway_api/src/search_api/models.py", line 142, in from_icat
    raise ValidationError(errors=[error_wrapper], model=cls)
[...]

PR #341, opened 2022-02-23
Fix requests which include parameters and filter on them

MRichards99:
Since there's another issue in #319 which hasn't been resolved (changing the response code), I think #319 should remain open.

## Description
This PR fixes the bug spotted in https://github.com/ral-facilities/datagateway-api/issues/319#issuecomment-1046890044. Initially, this bug was caused by the `Parameter` mapping - it would be sent to the Python ICAT `Query` class as a list rather than  a string for the entity name. To fix this, I've added parameters in `get_query_filter()`, `get_where_filter()` and `get_include_filter()` to pass around a new variable called `related_entity_name`. This is used to determine what goes into `entity_class_name` (another new variable) - see the ternary operators in `get_where_filter()` and `get_include_filter()` to see how this decision gets made.

As I fixed the inital bug, some of the query filter factory tests failed, so this PR is an aggregation of fixing the different things which I can across.

The version bump will be to 4.1.2 as this is a bug fix.

## Testing Instructions
The requests relating to this issue in #319 shouldn't error.

- [ ] Review code

### DIFF
--- a/datagateway_api/src/search_api/query_filter_factory.py
+++ b/datagateway_api/src/search_api/query_filter_factory.py
@@ -19,7 +19,7 @@ log = logging.getLogger()
 
 class SearchAPIQueryFilterFactory(QueryFilterFactory):
     @staticmethod
-    def get_query_filter(request_filter, entity_name=None):
+    def get_query_filter(request_filter, entity_name=None, related_entity_name=None):
         """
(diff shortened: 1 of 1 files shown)


VKTB (approved):
Thanks for sorting this out, glad it wasn't something more complicated.

PR #364, opened 2022-05-19
This fixes the problem with limit when it is not a number

### LINKED ISSUE #363: A filterError will be thrown if limit is string
**Description**:
The Panosc API accepts the limit filter to be a string therefore this is a valid query:
```
https://federated.scicat.ess.eu/api/documents?filter={"query":"diffraction","limit":"50"}
```

However, the current implementation does not accept limit to be a string (which I find safer but incompatible with the current PANOSC implementation) and:
```
/search-api/datasets?filter={"limit":"3"}
```
returns
```
{
  "error": {
    "message": "'>=' not supported between instances of 'str' and 'int'", 
    "name": "FilterError", 
    "statusCode": 400
  }
}
```

antolinos:
This PR will close #363 

## Description

## Testing Instructions
Both should work:
```
/search-api/datasets?filter={"limit":"3"}
/search-api/datasets?filter={"limit":3}
```

- [ ] Review code
- [ ] {more steps here}

### DIFF
--- a/datagateway_api/src/search_api/filters.py
+++ b/datagateway_api/src/search_api/filters.py
@@ -148,7 +148,7 @@ class SearchAPIWhereFilter(PythonICATWhereFilter):
 
 class SearchAPISkipFilter(PythonICATSkipFilter):
     def __init__(self, skip_value):
-        super().__init__(skip_value, filter_use="search_api")
+        super().__init__(int(skip_value), filter_use="search_api")
 
     def apply_filter(self, query):
         return super().apply_filter(query.icat_query.query)
@@ -156,7 +156,7 @@ class SearchAPISkipFilter(PythonICATSkipFilter):
 
 class SearchAPILimitFilter(PythonICATLimitFilter):
     def __init__(self, limit_value):
-        super().__init__(limit_value)
+        super().__init__(int(limit_value))
 
     def apply_filter(self, query):
         return super().apply_filter(query.icat_query.query)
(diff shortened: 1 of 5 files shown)


MRichards99 (changes requested):
Thanks for making the change, I've had a look and I've got a couple of comments:
- I'm happy with the implementation to support string inputs on the limit parameter. Would you be able to do the same thing for `skip` too please? They expect the same type of input so it makes sense to make the change to `skip` whilst we're addressing this issue.
- If you could add tests to `test_search_api_limit_filter.py`, that would be great. For `test_valid_limit_value()`, you could do a similar thing for what you've already done on `test_valid_limit_filter()`. For `test_invalid_limit_value()`, please could you add an invalid test case, such as an genuine string (e.g. `"test"`). Once you've made the change for skip, please add the same tests for that filter too.
- On this repo, we use automated semantic versioning to automatically release new versions. For this to work, we use Angular style commit messages. Have a read through the [API versioning](https://github.com/ral-facilities/datagateway-api#api-versioning) section of the README (and the associated links) which should give you a good idea what's required. You can also have a browse of previous commit messages too. If you could rebase your commits so you they follow this convention, that would be great. Usually there's a CI job that runs to check the commit messages automatically, but that doesn't appear to have run on this one. I suspect that it's because the branch is on a fork. Not to worry though, I can check them before merging them :)

MRichards99 (approved):
Brilliant, thanks for making those changes, the release should work now when this is merged. I quickly changed the exception that pytest looks for to a `ValueError` but it looks all good. Let me know if you want to make any further changes, othewise I will merge this tomorrow morning :)

MRichards99:
> I think that the `rebase` or `squash` will clean the previous commit messages and it can be done via github merging button, right?

It might do, I'm not actually sure. I've never used the rebase or squash buttons on the website, I will try to find out at some point out of curiosity. Thanks for making the changes, it looks good now.

PR #366, opened 2022-05-19
Draft: Support query filter and scoring

### LINKED ISSUE #365: Support query filter
**Description**:
Panosc API uses the query parameters as a valid filter and it's currently not supported by datagateway/search-api

Example of valid request:
```
https://federated.scicat.ess.eu/api/documents?filter={"query":"diffraction","limit":"50"}
```

Response from datagateway when `/search-api/datasets?filter={"query":"diffraction", "limit":"50"}`:
```
{
  "error": {
    "message": "No valid filter name given within filter query param: query", 
    "name": "FilterError", 
    "statusCode": 400
  }
}
```
This issue supersedes #357 

Query is uniquely used by the [scoring app](https://github.com/panosc-eu/panosc-search-scoring) which will calculate the score of the filtered results and then return the scored entities.

### Use cases:

Assuming that we want to make the use of the scoring optional then there are two cases

**1) With scoring**
If datagateway is configured to use the scoring app via a configuration parameter then the entities filtered should be scored. In practice, it means something like:

1. Search API filters the results
2. Results are sent to the scoring (that has been accordingly populated and the weights have been computed)
3. Aggregate the results of the scoring 

**2) Without scoring**
In case that scoring is not configured then I'd propose to:

1. Throw an error when query is present in the filter
4. Ignore the query parameter if not configured

antolinos:
This PR will close #365

## Description

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Connect to #{issue number}

### DIFF
--- a/datagateway_api/src/api_start_utils.py
+++ b/datagateway_api/src/api_start_utils.py
@@ -152,7 +152,8 @@ def create_api_endpoints(flask_app, api, specs):
     # DataGateway API endpoints
     if Config.config.datagateway_api is not None:
         datagateway_api_spec = next(
-            (spec for spec in specs if spec.title == "DataGateway API"), None,
+            (spec for spec in specs if spec.title == "DataGateway API"),
+            None,
         )
         try:
             backend_type = flask_app.config["TEST_BACKEND"]
@@ -223,7 +224,8 @@ def create_api_endpoints(flask_app, api, specs):
 
         # Session endpoint
         session_endpoint_resource = session_endpoints(
-            backend, client_pool=icat_client_pool,
+            backend,
+            client_pool=icat_client_pool,
         )
         api.add_resource(
             session_endpoint_resource,
@@ -234,7 +236,8 @@ def create_api_endpoints(flask_app, api, specs):
 
         # Table specific endpoints
         instrument_facility_cycle_resource = instrument_facility_cycles_endpoint(
-            backend, client_pool=icat_client_pool,
+            backend,
+            client_pool=icat_client_pool,
         )
         api.add_resource(
             instrument_facility_cycle_resource,
@@ -244,7 +247,8 @@ def create_api_endpoints(flask_app, api, specs):
         datagateway_api_spec.path(resource=instrument_facility_cycle_resource, api=api)
 
         count_instrument_facility_cycle_res = count_instrument_facility_cycles_endpoint(
-            backend, client_pool=icat_client_pool,
+            backend,
+            client_pool=icat_client_pool,
         )
         api.add_resource(
             count_instrument_facility_cycle_res,
@@ -254,7 +258,8 @@ def create_api_endpoints(flask_app, api, specs):
         datagateway_api_spec.path(resource=count_instrument_facility_cycle_res, api=api)
 
         instrument_investigation_resource = instrument_investigation_endpoint(
-            backend, client_pool=icat_client_pool,
+            backend,
+            client_pool=icat_client_pool,
         )
         api.add_resource(
             instrument_investigation_resource,
@@ -265,7 +270,8 @@ def create_api_endpoints(flask_app, api, specs):
         datagateway_api_spec.path(resource=instrument_investigation_resource, api=api)
 
         count_instrument_investigation_res = count_instrument_investigation_endpoint(
-            backend, client_pool=icat_client_pool,
+            backend,
+            client_pool=icat_client_pool,
         )
         api.add_resource(
             count_instrument_investigation_res,
@@ -283,13 +289,14 @@ def create_api_endpoints(flask_app, api, specs):
     # Search API endpoints
     if Config.config.search_api is not None:
         search_api_spec = next(
-            (spec for spec in specs if spec.title == "Search API"), None,
+            (spec for spec in specs if spec.title == "Search API"),
+            None,
         )
         search_api_extension = Config.config.search_api.extension
         search_api_entity_endpoints = {
-            "datasets": "Dataset",
-            "documents": "Document",
-            "instruments": "Instrument",
+            "Datasets": "Dataset",
+            "Documents": "Document",
+            "Instruments": "Instrument",
         }
 
         for endpoint_name, entity_name in search_api_entity_endpoints.items():
--- a/datagateway_api/src/api_start_utils.py
+++ b/datagateway_api/src/api_start_utils.py
@@ -152,8 +152,7 @@ def create_api_endpoints(flask_app, api, specs):
     # DataGateway API endpoints
     if Config.config.datagateway_api is not None:
         datagateway_api_spec = next(
-            (spec for spec in specs if spec.title == "DataGateway API"),
-            None,
+            (spec for spec in specs if spec.title == "DataGateway API"), None,
         )
         try:
             backend_type = flask_app.config["TEST_BACKEND"]
@@ -224,8 +223,7 @@ def create_api_endpoints(flask_app, api, specs):
 
         # Session endpoint
         session_endpoint_resource = session_endpoints(
-            backend,
-            client_pool=icat_client_pool,
+            backend, client_pool=icat_client_pool,
         )
         api.add_resource(
             session_endpoint_resource,
@@ -236,8 +234,7 @@ def create_api_endpoints(flask_app, api, specs):
 
         # Table specific endpoints
         instrument_facility_cycle_resource = instrument_facility_cycles_endpoint(
-            backend,
-            client_pool=icat_client_pool,
+            backend, client_pool=icat_client_pool,
         )
         api.add_resource(
             instrument_facility_cycle_resource,
@@ -247,8 +244,7 @@ def create_api_endpoints(flask_app, api, specs):
         datagateway_api_spec.path(resource=instrument_facility_cycle_resource, api=api)
 
         count_instrument_facility_cycle_res = count_instrument_facility_cycles_endpoint(
-            backend,
-            client_pool=icat_client_pool,
+            backend, client_pool=icat_client_pool,
         )
         api.add_resource(
             count_instrument_facility_cycle_res,
@@ -258,8 +254,7 @@ def create_api_endpoints(flask_app, api, specs):
         datagateway_api_spec.path(resource=count_instrument_facility_cycle_res, api=api)
 
         instrument_investigation_resource = instrument_investigation_endpoint(
-            backend,
-            client_pool=icat_client_pool,
+            backend, client_pool=icat_client_pool,
         )
         api.add_resource(
             instrument_investigation_resource,
@@ -270,8 +265,7 @@ def create_api_endpoints(flask_app, api, specs):
         datagateway_api_spec.path(resource=instrument_investigation_resource, api=api)
 
         count_instrument_investigation_res = count_instrument_investigation_endpoint(
-            backend,
-            client_pool=icat_client_pool,
+            backend, client_pool=icat_client_pool,
         )
         api.add_resource(
             count_instrument_investigation_res,
@@ -289,8 +283,7 @@ def create_api_endpoints(flask_app, api, specs):
     # Search API endpoints
     if Config.config.search_api is not None:
         search_api_spec = next(
-            (spec for spec in specs if spec.title == "Search API"),
-            None,
+            (spec for spec in specs if spec.title == "Search API"), None,
         )
         search_api_extension = Config.config.search_api.extension
         search_api_entity_endpoints = {
--- a/datagateway_api/src/common/config.py
+++ b/datagateway_api/src/common/config.py
@@ -136,7 +136,7 @@ class SearchAPI(BaseModel):
     scoring_enabled: StrictBool
     scoring_server: StrictStr
     scoring_group: StrictStr
-    scoring_limit : StrictInt
+    scoring_limit: StrictInt
 
     _validate_extension = validator("extension", allow_reuse=True)(validate_extension)
 
--- a/datagateway_api/src/common/config.py
+++ b/datagateway_api/src/common/config.py
@@ -136,6 +136,7 @@ class SearchAPI(BaseModel):
     scoring_enabled: StrictBool
     scoring_server: StrictStr
     scoring_group: StrictStr
+    scoring_limit : StrictInt
 
     _validate_extension = validator("extension", allow_reuse=True)(validate_extension)
 
--- a/datagateway_api/src/common/config.py
+++ b/datagateway_api/src/common/config.py
@@ -133,6 +133,9 @@ class SearchAPI(BaseModel):
     mechanism: StrictStr
     username: StrictStr
     password: StrictStr
+    scoring_enabled: StrictBool
+    scoring_server: StrictStr
+    scoring_group: StrictStr
 
     _validate_extension = validator("extension", allow_reuse=True)(validate_extension)
 
--- a/datagateway_api/src/datagateway_api/icat/filters.py
+++ b/datagateway_api/src/datagateway_api/icat/filters.py
@@ -9,8 +9,7 @@ from datagateway_api.src.common.filters import (
     OrderFilter,
     ScoringQueryFilter,
     SkipFilter,
-    WhereFilter,
-    QueryFilter
+    WhereFilter
 )
 from datagateway_api.src.common.helpers import get_icat_properties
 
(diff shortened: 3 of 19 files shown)


MRichards99:
I can see this is a draft so I haven't looked at these changes in depth, but if you could also rebase your commits to follow the Angular commit message convention as described in https://github.com/ral-facilities/datagateway-api/pull/364#pullrequestreview-981835852, that would be great

datagateway_api/src/api_start_utils.py
```diff
@@ -152,7 +152,8 @@ def create_api_endpoints(flask_app, api, specs):
     # DataGateway API endpoints
     if Config.config.datagateway_api is not None:
         datagateway_api_spec = next(
-            (spec for spec in specs if spec.title == "DataGateway API"), None,
+            (spec for spec in specs if spec.title == "DataGateway API"),
```
MRichards99, line 155: There's been some code formatting changes where my dev environment changes them back to the way they used to be. I assume this is because of a difference in Black versions. When running `nox -s black`, the latest version is installed. I will look into how I can (cleanly) fix a version when running that command. There's nothing for you to do here, but this serves as a reminder for me.

datagateway_api/src/api_start_utils.py
```diff
@@ -283,13 +289,14 @@ def create_api_endpoints(flask_app, api, specs):
...
+            None,
         )
         search_api_extension = Config.config.search_api.extension
         search_api_entity_endpoints = {
-            "datasets": "Dataset",
-            "documents": "Document",
-            "instruments": "Instrument",
+            "Datasets": "Dataset",
+            "Documents": "Document",
+            "Instruments": "Instrument",
```
MRichards99, line 299: Do you know why this has been changed? The keys are for the endpoints, which are typically lower case

datagateway_api/src/common/config.py
```diff
@@ -133,6 +133,9 @@ class SearchAPI(BaseModel):
     mechanism: StrictStr
     username: StrictStr
     password: StrictStr
+    scoring_enabled: StrictBool
+    scoring_server: StrictStr
+    scoring_group: StrictStr
```
MRichards99, line 138: Could you add example values to `config.json.example` please? Keeping the scoring disabled might be best in case of someone using DataGateway API only, not the search API.

datagateway_api/src/datagateway_api/icat/filters.py
```diff
@@ -280,6 +281,14 @@ def icat_set_limit(query, skip_number, limit_number):
...
 
 
+class PythonICATQueryFilter(ScoringQueryFilter):
+    def __init__(self, value):
+        super().__init__(value)
+
+    def apply_filter(self, query):
+        return
+
+
```
MRichards99, line 291: This class isn't needed because it features no implementation. When defining the filter in `search_api/filters.py`, you can inherit directly from `ScoringQueryFilter`.

datagateway_api/src/resources/search_api_endpoints.py
```diff
@@ -31,7 +34,27 @@ class Endpoint(Resource):
...
+            if not is_query_parameter_enabled(filters):
+                entities = get_search(entity_name, filters)
+                return entities, 200
+            else:
+                query = get_search_api_query_filter_list(filters)[0].value
+
+                entities = get_search(
+                    entity_name,
+                    filters,
+                    "LOWER(o.summary) like '%" + query.lower() + "%'",
```
MRichards99, line 48: Looking at this line and line 56 (where you pass `"investigations"` to `get_score()`), I guess this only works on /documents for now? Have you got a plan of how to make it work for the other endpoints?

datagateway_api/src/resources/search_api_endpoints.py
```diff
@@ -31,7 +34,27 @@ class Endpoint(Resource):
...
+                )
+                log.debug(
+                    "Applying score to %s entities with query %s",
+                    len(entities),
+                    query,
+                )
+                log.debug("Entities retrieved %s", len(entities))
```
MRichards99, line 55: `len(entities)` is logged in the log statement above, so this one could be removed

datagateway_api/src/search_api/filters.py
```diff
@@ -52,8 +53,10 @@ def apply_filter(self, query):
 
         # Convert PaNOSC field names to ICAT field names
         for field_name in panosc_field_names:
-            panosc_mapping_name, icat_field_name = mappings.get_icat_mapping(
-                panosc_mapping_name, field_name,
+            panosc_mapping_name, icat_field_name = mappings.get_icat_mapping
+            (
+                panosc_mapping_name,
+                field_name,
             )
```
MRichards99, line 60: Note for me: formatting change within a file that has functionality changes

datagateway_api/src/search_api/filters.py
```diff
@@ -162,6 +165,14 @@ def apply_filter(self, query):
         return super().apply_filter(query.icat_query.query)
 
 
+class SearchAPIQueryFilter(PythonICATQueryFilter):
```
MRichards99, line 168: As per my previous comment, you can inherit directly from `ScoringQueryFilter`. Would you be able to rename this class to `SearchAPIScoringFilter` please? When searching for `SearchAPIQueryFilter`, it also returns results for `SearchAPIQueryFilterFactory`, unless you have a match whole word filter set.

datagateway_api/src/search_api/helpers.py
```diff
@@ -74,8 +84,76 @@ def create_error_message(e):
     return wrapper_error_handling
 
 
+def get_score(entities, query, group, limit):
```
MRichards99, line 87: I can see this function is yet to be cleaned up, so I won't review it too much. I believe the purpose of this function to to get the scoresfrom the scoring server and then add them to the entities? Would you be able to split this functionality into two functions?

Keep the request to the scoring server in this function and return the scores. Then have a second function (called `add_scores_to_entities()` or something similar), pass in `entities` and `scores`, then add the scores to the entities.

Implementing it using two functions ensures a single function has a single purpose, and doesn't try to do many things at once

datagateway_api/src/search_api/helpers.py
```diff
@@ -74,8 +84,76 @@ def create_error_message(e):
...
+    """
+    Gets the score on the given entities based in the query parameter and limit
+    """
```
MRichards99, line 90: This docstring needs more details, including `:param:` and `:type:` of each parameter, following the same conventions as other docstrings in this repo

datagateway_api/src/search_api/helpers.py
```diff
@@ -74,8 +84,76 @@ def create_error_message(e):
...
+
+    try:
+        myobj = {
+            "query": query,
+            "group": Config.config.search_api.scoring_group,
+            "limit": limit,
+            "itemIds": list(map(lambda entity: entity["pid"], entities)),
+        }
+        response = requests.post(
+            Config.config.search_api.scoring_server, json=myobj, timeout=5
```
MRichards99, line 100: I like the use of `timeout=`, so requests won't get stuck if the scoring server isn't working

datagateway_api/src/search_api/helpers.py
```diff
@@ -74,8 +84,76 @@ def create_error_message(e):
...
+                # )["score"]
+            return entities
+        else:
+            raise ScoringAPIError(
+                Exception(f"Score API returned {response.status_code}"),
+            )
+    except ConnectionError as e:
+        log.error("ConnectionError to %s ", url)
+        raise e
+    except Exception as e:
```
MRichards99, line 149: Are there any other specific exceptions that we could catch, instead of a generic `Exception`?

datagateway_api/src/search_api/helpers.py
```diff
@@ -198,8 +278,11 @@ def get_files(entity_name, pid, filters):
     """
 
     log.info("Getting files of dataset (PID: %s), using request's filters", pid)
-    log.debug(
-        "Entity Name: %s, Filters: %s", entity_name, filters,
+    log.debug
+    (
+        "Entity Name: %s, Filters: %s",
+        entity_name,
+        filters,
```
MRichards99, line 285: Note for me: formatting change within a file that has functionality changes

datagateway_api/src/search_api/helpers.py
```diff
@@ -221,11 +304,31 @@ def get_files_count(entity_name, filters, pid):
...
+    return list(filter(lambda x: isinstance(x, SearchAPIQueryFilter), filters))
+
+
+@client_manager
+def is_query_parameter_enabled(filters):
+    """
+    Checks if query parameter is enabled based if the SearchAPIQueryFilter
+    is in the query and the scoring is enabled
+    """
+    print(Config.config.search_api.password)
```
MRichards99, line 333: I know this is for testing, but this `print()` should be removed

datagateway_api/src/search_api/panosc_mappings.py
```diff
@@ -43,14 +44,14 @@ def get_icat_mapping(self, panosc_entity_name, field_name):
...
-        )
+        # log.debug(
+        #    "Searching mapping file to find ICAT translation for %s",
+        #    f"{panosc_entity_name}.{field_name}",
+        # )
 
         try:
             icat_mapping = self.mappings[panosc_entity_name][field_name]
-            log.debug("ICAT mapping/translation found: %s", icat_mapping)
+            # log.debug("ICAT mapping/translation found: %s", icat_mapping)
```
MRichards99, line 54: Was this a bit verbose? I can't remember if I found this too noisy when I ran the API. If so, feel free to remove this line (and other logging statement above) completely

datagateway_api/src/search_api/helpers.py
```diff
@@ -74,8 +84,76 @@ def create_error_message(e):
...
+                # filter(
+                #    lambda x: x["itemId"] == entity["pid"], response.json()
+                # )["score"]
+            return entities
+        else:
+            raise ScoringAPIError(
+                Exception(f"Score API returned {response.status_code}"),
+            )
+    except ConnectionError as e:
+        log.error("ConnectionError to %s ", url)
```
MRichards99, line 147: `url` is an undefined variable

datagateway_api/src/resources/search_api_endpoints.py
```diff
@@ -31,7 +34,27 @@ class Endpoint(Resource):
         def get(self):
             filters = get_filters_from_query_string("search_api", entity_name)
             log.debug("Filters: %s", filters)
-            return get_search(entity_name, filters), 200
+            log.debug("entity_name: %s", entity_name)
```
MRichards99, line 37: This could be removed, or combined with the logging statement above

datagateway_api/src/search_api/helpers.py
```diff
@@ -74,8 +84,76 @@ def create_error_message(e):
...
+        response = requests.post(
+            Config.config.search_api.scoring_server, json=myobj, timeout=5
+        )
+
+        if response.status_code < 400:
+            scores = response.json()["scores"]
+            print(str(len(scores)) + " scores retrieved from scoring app")
+            print(str(len(entities)) + " entities retrieved from ICAT")
+            for entity in entities:
+                items = list(
```
MRichards99, line 108: I'm unsure of how the JSON response will look so I can't immediately comment on this logic. When I test the changes, I'll come back to this comment and give further thoughts

datagateway_api/src/search_api/filters.py
```diff
@@ -180,8 +191,10 @@ def apply_filter(self, query):
             split_panosc_fields = panosc_field_name.split(".")
 
             for split_field in split_panosc_fields:
-                panosc_entity_name, icat_field_name = mappings.get_icat_mapping(
-                    panosc_entity_name, split_field,
+                panosc_entity_name, icat_field_name = mappings.get_icat_mapping
+                (
+                    panosc_entity_name,
+                    split_field,
```
MRichards99, line 197: Note for me: formatting change within a file that has functionality changes

datagateway_api/src/api_start_utils.py
antolinos: I needed to change that in order to match the PANOSC endpoints that use upper case:
https://github.com/panosc-eu/search-api/blob/master/doc/dataset-example-queries.md#curl

Otherwise, the aggregator will return 404

datagateway_api/src/common/config.py
antolinos: This is my config:
```
    "scoring_enabled": true,
    "scoring_server": "http://dau-dm-01:9000/score?limit=2000",
    "scoring_group": "investigation"
```

datagateway_api/src/resources/search_api_endpoints.py
antolinos: You are right and honestly I do not know. There are several reason:
1) I focused on making https://data.panosc.eu/ to work. Surprisingly I did discover that only use `/documents` endpoint. So, I did not consider to add something that is not used
2) When I tried to use the scoring app with the datasets, it did not work when calculating the score (my guess is because of the number of datasets). I created an issue: 
https://github.com/panosc-eu/panosc-search-scoring/issues/9

So, yes, it might be needed in the future but it very uncertain and first we would need the scoring app to work at the level of datasets

MRichards99: We were slightly concerned about the way the score calculation behaves and that it might cause performance issues with large volumes of data. It would be good to confirm (from someone who wrote the scoring software perhaps?) which endpoint(s) the scoring needs to work on.

datagateway_api/src/common/config.py
MRichards99: I guess this is your own scoring server? Does this mean I'll need to setup my own instance to test your changes? Is there a PaNOSC scoring server that I could use?

datagateway_api/src/api_start_utils.py
MRichards99: Ah ok, I didn't notice that before but I guess that's a required change if the server on PaNOSC's side won't recognise lowercase endpoint names.

antolinos: Exact.

datagateway_api/src/common/config.py
antolinos: No, there is not as far as I know and take into account that you need to install and populate with your own data in order to calculate the weights. 
It is not a big deal but has to be done.

datagateway_api/src/search_api/helpers.py
antolinos: Not sure to understand, there's a timeout no?
timeout=5  

or what do you mean?

datagateway_api/src/search_api/panosc_mappings.py
antolinos: I presume that it was too verbose but I do not mind to put it back

datagateway_api/src/search_api/helpers.py
MRichards99: Ah, a typo from me but I was complimenting the use of timeouts here :)

datagateway_api/src/search_api/panosc_mappings.py
MRichards99: No you can remove this line. I do remember it was verbose when the line was added so I think this can be removed.

datagateway_api/src/common/config.py
MRichards99: That's good to know, I will setup my own and test this branch when I get a chance

MRichards99:
> @MRichards99 There are some linting errors with a unclear error message. What does it mean `BLK100 Black would make changes.`?
> 
> I understand and share your concerns about the performance, that is why I would be happy to keep the scoring at the level of the investigations only for the moment. Let me know what needs to be changed.
> 
> Cheers, Alex

@antolinos this error message is from flake8 saying that the code doesn't meet the code formatting standards set out by the formatter [black](https://black.readthedocs.io/en/stable/). If you have the Nox sessions working, you can run `nox -s black` to get Black to correctly format the code (it will make changes automatically). It only takes a few seconds to run so if you're not using the nox sessions, I would be happy to run the formatter on your branch and push the changes. Just let me know if/when you'd like me to do that?

antolinos:
@MRichards99 

Thanks for the info. I ran black and lint so the format should be fine.  However, lot of problems starting by the db_generator.
Do you know what am I missing?

VKTB:
Superseded by https://github.com/ral-facilities/datagateway-api/pull/399

PR #393, opened 2022-11-30
Tests/reorganise tests#375

### LINKED ISSUE #375: Improve DataGateway API Tests
**Description**:
The tests for this repository have relatively high coverage but the types of tests are mixed up - unit tests and integration tests are together. So long as the person/machine running the tests has an ICAT stack running with some LILS facility data then that's fine but I'm not sure we should always assume that to run the unit tests.

I haven't looked into the best way to arrange tests but doing this (and drawing on any past knowledge) would be a good place to start. In an ideal world, I think we should be able to run unit tests and integration tests separately - I've seen ICAT repos actually split the tests up into `unit/` and `integration/` directories but I would be open to alternative solutions if we see they would work well (flags on tests within Pytest maybe?). 

**Acceptance criteria**:
- [ ] Arrange tests more logically so that we can run unit tests (that don't assume a running ICAT stack) on their own

Reillyhewitson:
This PR will close #375 

## Description
This PR reorganises the tests into integration tests that require an ICAT stack to be running and unit tests that don't.

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Connect to #375


Reillyhewitson:
> Is there a command that can be used to only run the unit tests and another that will only run the integration tests?

Yes, that would be possible, with pytest you can point it at a directory and it will run all the tests in that directory and any directories below. I'm not so sure that if using that to decide whether or not to run integration testing would be possible with our current actions setup, it might be something that needs to be done with composite actions, but I can look into it.

@VKTB

VKTB:
> It would also be possible to set up a workflow to setup and run integration tests once unit tests have passed.

I don't think we need to set up a new workflow. We can just rename the existing `tests` job in the existing [ci-build.yml](https://github.com/ral-facilities/datagateway-api/blob/71d81ac444be2cf2ff18fae1eb7f5108a819d55b/.github/workflows/ci-build.yml#L9-L129) workflow to `integration-tests` (or something along those lines) since that already sets up an ICAT stack and populates it with dummy data, and then just define a `unit-tests` job before it that will only run the unit tests.

MRichards99 (changes requested):
All the changes you've made look good, would you be able to separate the tests as you and Viktor have discussed please? Just put them into two separate jobs (in the existing workflow file) but have them run parallel like all the workflows do currently. I'm not sure I can see a reason where the integration tests need to wait until the unit tests have passed (happy to be persuaded otherwise though?)

VKTB:
>I'm not sure I can see a reason where the integration tests need to wait until the unit tests have passed (happy to be persuaded otherwise though?)

In my opinion, I think that the job that runs the integration tests should only start if the job that runs the unit tests passes so not in parallel. This is because if any of the unit tests fail then the chances of some integration tests failing are high (if not 100%). If we run them in parallel then we will just end up having to wait (+/- 15 minutes) for the integrations test job to run while knowing that it will likely fail.

MRichards99 (changes requested):
Looks good, the split is great. I've just realised some of the README needs updating to reflect the changes you've made (i.e. the name of the Nox sessions has changed). Could you update [Automated Testing & Other Development Helpers (Nox)](https://github.com/ral-facilities/datagateway-api/tree/tests/reorganise-tests%23375#automated-testing--other-development-helpers-nox) and [Running Tests](https://github.com/ral-facilities/datagateway-api/tree/tests/reorganise-tests%23375#running-tests) please?

MRichards99 (approved):

VKTB:
I was just looking at the GitHub Action logs to see how long the unit tests take to run and wondered why they take ~1m45s. After looking at it more closely, I realised it is because the `unit_tests` `nox` session runs the `pip uninstall -y setuptools`, `pip install setuptools<58.0.0` and `poetry install` commands. Do you think that we can safely remove the [_Uninstall setuptools_, _Install older setuptools_ and _Install dependencies_](https://github.com/ral-facilities/datagateway-api/blob/aa4350c3b9c8e95cfdb3cdb5465f5dba958f8373/.github/workflows/ci-build.yml#L41-L56) steps in the `tests` CI job given that the `nox` session takes care of this? The `integration_tests` `nox` session also does the same so the steps that I mentioned above seem redundant to me.

MRichards99 (changes requested):
As per comment

MRichards99 (dismissed):

MRichards99 (approved):

PR #399, opened 2023-01-25
Query filter and scoring

antolinos:

## Description

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Connect to #{issue number}

### DIFF
--- a/datagateway_api/config.yaml.example
+++ b/datagateway_api/config.yaml.example
@@ -15,6 +15,10 @@ search_api:
   mechanism: "anon"
   username: ""
   password: ""
+  scoring_enabled: false,
+  scoring_server: "http://localhost:9000/score",
+  scoring_group: "investigation" #corresponds to the defined group in the scoring app. https://github.com/panosc-eu/panosc-search-scoring/blob/master/docs/md/PaNOSC_Federated_Search_Results_Scoring_API.md#model
+  scoring_limit: 1000
 flask_reloader: false
 log_level: "DEBUG"
 log_location: "/home/runner/work/datagateway/datagateway/datagateway-api/datagateway_api/logs.log"


datagateway_api/config.yaml.example
```diff
@@ -15,6 +15,10 @@ search_api:
   mechanism: "anon"
   username: ""
   password: ""
+  scoring_enabled: false,
+  scoring_server: "http://localhost:9000/score",
+  scoring_group: "investigation" #corresponds to the defined group in the scoring app. https://github.com/panosc-eu/panosc-search-scoring/blob/master/docs/md/PaNOSC_Federated_Search_Results_Scoring_API.md#model
+  scoring_limit: 1000
```
VKTB, line 21:
```suggestion
  scoring_enabled: false
  scoring_server: "http://localhost:9000/score"
  scoring_group: "investigation" #corresponds to the defined group in the scoring app. https://github.com/panosc-eu/panosc-search-scoring/blob/master/docs/md/PaNOSC_Federated_Search_Results_Scoring_API.md#model
  scoring_limit: 1000
```

datagateway_api/src/search_api/panosc_mappings.py
```diff
@@ -50,7 +50,7 @@ def get_icat_mapping(self, panosc_entity_name, field_name):
 
         try:
             icat_mapping = self.mappings[panosc_entity_name][field_name]
-            log.debug("ICAT mapping/translation found: %s", icat_mapping)
+            # Too verbose log.debug("ICAT mapping/translation found: %s", icat_mapping)
```
VKTB, line 53: I agree so best to just remove it entirely.
```suggestion
```

datagateway_api/src/search_api/helpers.py
```diff
@@ -74,8 +78,69 @@ def create_error_message(e):
...
+        else:
+            raise ScoringAPIError(
+                Exception(f"Score API returned {response.status_code}"),
+            )
+    except ValueError as e:
+        log.error("Response is not a valid json")
+        raise e
+    except ConnectionError as e:
+        log.error("ConnectionError to %s ", Config.config.search_api.scoring_server)
+        raise e
```
VKTB, line 114: This exception is not caught at a higher level as part of the `search_api_error_handling` decorator method, meaning that the whole stack trace is returned to the consumer. We don't want to do this as it is considered bad security practice. It might be best if we raise `ScoringAPIError` here and also in the places where the `ValueError` and `Exception` are caught.

VKTB (changes requested):

datagateway_api/src/search_api/helpers.py
```diff
@@ -74,8 +78,69 @@ def create_error_message(e):
...
+    :param entities: List of entities that have been retrieved from one ICAT query.
+    :type entities: :class:`list`
+    :param scores: List of items retrieved from the scoring application
+    :type scores: :class:`list`
+    """
+    for entity in entities:
+        entity["score"] = -1
+        items = list(
+            filter(
+                lambda score: f"pid:{score['itemId']}" == str(entity["pid"]), scores,
```
VKTB, line 134: Not sure why `score` `itemId`s are prefixed with `pid:` here but not all documents are prefixed with `pid:` by the Search API. For example, if the ICAT investigation has a `doi` then the `doi` is used as `pid` else its `id` is prefixed with `pid:`. This is exactly the reason why I was getting `-1` score for one of the documents that had a `doi` as its `pid`.

To ensure that the `pid`s are consistent across the Search API and the scoring API, I think that it's best to use the Search API when retrieving the documents/ investigations from ICAT so that they get prefixed with `pid:` where necessary. The results can then be passed to the scoring app. Let me know if this is not making sense and we can have a chat.

VKTB (changes requested):

datagateway_api/src/resources/search_api_endpoints.py
```diff
@@ -30,8 +35,21 @@ class Endpoint(Resource):
...
-            return get_search(entity_name, filters), 200
+            # in case there is no query filter then we processed as usual
+            if not is_query_filter(filters):
+                return get_search(entity_name, filters), 200
+            else:
+                query = get_search_api_query_filter_list(filters)[0].value
+                entities = get_search(
+                    entity_name,
+                    filters,
+                    "LOWER(o.summary) like '%" + query.lower() + "%'",
```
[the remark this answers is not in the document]
antolinos: Yes, much nicer

antolinos:
Dear @VKTB 

It is all good from my side. Thank you very much!

VKTB:
Thank you too @antolinos!

VKTB:
The failing tests will be fixed by the [Fix/remove dataset filesize #403](https://github.com/ral-facilities/datagateway-api/pull/403) PR.

VKTB (approved):

PR #409, opened 2023-03-29
Fix swagger docs at non-root paths #408

### LINKED ISSUE #408: Deployment in Apache with mod_wsgi: deploy at non-root path
**Description**:
`datagateway-api` works fine when deployed in Apache with [mod_wsgi](https://modwsgi.readthedocs.io/en/master/), _if_ it is deployed at the document root.

E.g., if you have something like the following:

`httpd.conf`:
```xml
<VirtualHost _default_:443>

        ServerName data.example.org:443

        WSGIDaemonProcess datapub \
                display-name=%{GROUP} user=wwwrun group=www threads=3
        WSGIScriptAlias / /usr/lib/python3.6/site-packages/datagateway_api/wsgi.py \
                process-group=datapub application-group=%{GLOBAL}

</VirtualHost>
```
`config.yaml`:
```yaml
search_api:
  extension: "/search-api"
```

The search API works and is reachable at `https://data.example.org/search-api`.

However, this is very inconvenient, as it shadows the whole URL tree, because any request will be absorbed by the WSGI application.  Nothing else besides `/search-api` is reachable at `https://data.example.org/` anymore, unless it is explicitely aliased.

For HZB this is also a problem, because I normally want to have a redirection from `https://data.helmholtz-berlin.de/` to the ICAT web UI at `https://topcat.hemlholtz-berlin.de/`. This redirection is now lost after I deployed the search API as described above at `https://data.helmholtz-berlin.de/search-api`. I can't resurrect that redirection with an explicit alias, because I can't redirect `/` at the same time as having a WSGI script alias at `/`.

Obviously, what I would like to do is that only requests to the search API will be handled by the WSGI script.  E.g., I would like to configure something like
```xml
<VirtualHost _default_:443>

        ServerName data.example.org:443

        WSGIDaemonProcess datapub \
                display-name=%{GROUP} user=wwwrun group=www threads=3
        WSGIScriptAlias /search-api /usr/lib/python3.6/site-packages/datagateway_api/wsgi.py \
                process-group=datapub application-group=%{GLOBAL}

</VirtualHost>
```
in the `httpd.conf`.

But that doesn't seem to work.

If I try this with the same `config.yaml` as above, I get a `404 Not found` at `https://data.example.org/search-api`. The search API is apparently listening at `https://data.example.org/search-api/search-api`, but it does not work properly, because Swagger tries to load the API definition from `https://data.example.org/search-api/openapi.json`, which fails. It is available at `https://data.example.org/search-api/search-api/openapi.json` instead.

If I try to deploy the WSGI at `/search-api` as above and set the following in the `config.yaml`:
```yaml
search_api:
  extension: ""
```
Then the search API is again apparently listening at `https://data.example.org/search-api`, but I get just a blank page.  The reason becomes obvious obvious if I look into the HTML source:
```html
<!-- HTML for static distribution bundle build -->
<!DOCTYPE html>
<html lang="en">

<head>
  <meta charset="UTF-8">
  <title>Search API OpenAPI Spec</title>
  <link rel="stylesheet" type="text/css" href="/index.css">
  <link rel="stylesheet" type="text/css" href="/swagger-ui.css">
  <link rel="icon" type="image/png" href="/favicon-32x32.png" sizes="32x32" />
  <link rel="icon" type="image/png" href="/favicon-16x16.png" sizes="16x16" />
</head>

<body>
  <div id="swagger-ui"></div>

  <script src="/swagger-ui-bundle.js"> </script>
  <script src="/swagger-ui-standalone-preset.js"> </script>
  <!-- ... -->
</body>

</html>
```
It tries to load its components (CSS and JS) from `/...` instead of `/search-api/...`, which obviously must fail.
Essentially the same happens if I put `extension: "/"` in the `config.yaml`.

**Acceptance criteria**:
- [ ] get deploying of the search API using ` WSGIScriptAlias /search-api ...` to work.

louise-davies:
This PR will close #408

## Description
Add a url_prefix config item which tells the app where it's expected to be mounted. I've added some code to take url_prefix into consideration in development, but otherwise we'd be expecting this to be done via a WSGI server e.g. Apache mod_wsgi

This then fixes the swagger UI as the UI files & openapi.json have the correct URLs. Every other URL in the app already worked fine with aliases as they're handled as part of WSGI spec

## Testing Instructions

- [ ] Review code
- [ ] Run API locally with `url_prefix` option and set one of dg-api or search-api's aliases to "/" - test that the swagger UI is visible for both APIs
- [ ] Ideally also test with Apache mod_wsgi - or I can do a call with you and show you on my machine :)

## Agile Board Tracking
Closes #408

### DIFF
--- a/datagateway_api/config.yaml.example
+++ b/datagateway_api/config.yaml.example
@@ -30,3 +30,4 @@ host: "127.0.0.1"
 port: "5000"
 test_user_credentials: { username: "root", password: "pw" }
 test_mechanism: "simple"
+url_prefix: "/"
--- a/datagateway_api/src/api_start_utils.py
+++ b/datagateway_api/src/api_start_utils.py
@@ -69,8 +69,8 @@ class CustomErrorHandledApi(Api):
 
 def configure_datagateway_api_swaggerui_blueprint(flask_app):
     swaggerui_blueprint = get_swaggerui_blueprint(
-        base_url=Config.config.datagateway_api.extension,
-        api_url="/datagateway-api/openapi.json",
+        base_url=f"{Config.config.url_prefix}{Config.config.datagateway_api.extension}",
+        api_url=f"{Config.config.url_prefix}/datagateway-api/openapi.json",
         config={"app_name": "DataGateway API OpenAPI Spec"},
         blueprint_name="DataGateway API Swagger UI",
     )
@@ -83,8 +83,8 @@ def configure_datagateway_api_swaggerui_blueprint(flask_app):
 
 def configure_search_api_swaggerui_blueprint(flask_app):
     swaggerui_blueprint = get_swaggerui_blueprint(
-        base_url=Config.config.search_api.extension,
-        api_url="/search-api/openapi.json",
+        base_url=f"{Config.config.url_prefix}{Config.config.search_api.extension}",
+        api_url=f"{Config.config.url_prefix}/search-api/openapi.json",
         config={"app_name": "Search API OpenAPI Spec"},
         blueprint_name="Search API Swagger UI",
     )
--- a/datagateway_api/src/common/config.py
+++ b/datagateway_api/src/common/config.py
@@ -57,6 +57,9 @@ class DataGatewayAPI(BaseModel):
 
     _validate_extension = validator("extension", allow_reuse=True)(validate_extension)
 
+    def __getitem__(self, item):
+        return getattr(self, item)
+
     @validator("db_url", always=True)
     def require_db_config_value(cls, value, values):  # noqa: B902, N805
         """
@@ -147,6 +150,9 @@ class SearchAPI(BaseModel):
 
     _validate_extension = validator("extension", allow_reuse=True)(validate_extension)
 
+    def __getitem__(self, item):
+        return getattr(self, item)
+
 
 class TestUserCredentials(BaseModel):
     username: StrictStr
@@ -188,6 +194,9 @@ class APIConfig(BaseModel):
     url_prefix: StrictStr
 
     _validate_extension = validator("url_prefix", allow_reuse=True)(validate_extension)
+
+    def __getitem__(self, item):
+        return getattr(self, item)
 
     @classmethod
     def load(cls, path=Path(__file__).parent.parent.parent / "config.yaml"):
--- a/datagateway_api/src/common/config.py
+++ b/datagateway_api/src/common/config.py
@@ -32,6 +32,8 @@ def validate_extension(extension):
             raise ValueError("must start with '/'")
         if extension.endswith("/") and len(extension) != 1:
             raise ValueError("must not end with '/'")
+        if extension == "/":
+            extension = ""
 
     return extension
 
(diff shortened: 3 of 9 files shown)


MRichards99 (dismissed):

VKTB:
Sorry, I am late seeing this PR but I just noticed that it is introducing a new config value. From a versioning perspective, I think that this should be a breaking change (major version bump) rather than a fix (patch version bump).

MRichards99:
@VKTB ah yes, I missed that (it looks like the Semantic Release workflow thing doesn't run any more either). I think rewording [`26a901e758c1f45dd0b747fae049e7a0d32c9b89`](https://github.com/ral-facilities/datagateway-api/pull/409/commits/26a901e758c1f45dd0b747fae049e7a0d32c9b89) so it causes a bump to 8.0.0 would be the best thing to do? I've made the change locally so let me know if you agree and I can just force-push

VKTB:
> I think rewording [`26a901e758c1f45dd0b747fae049e7a0d32c9b89`](https://github.com/ral-facilities/datagateway-api/pull/409/commits/26a901e758c1f45dd0b747fae049e7a0d32c9b89) so it causes a bump to 8.0.0 would be the best thing to do? I've made the change locally so let me know if you agree and I can just force-push

@MRichards99 Yes that should do the job. While at it, could you also please merge `main` into this branch as it is behind?

> it looks like the Semantic Release workflow thing doesn't run any more either
From its [repo README](https://github.com/zeke/semantic-pull-requests), it Looks like it is no longer maintained and there was a security incident. It does mention some alternatives but we to make sure that they are secure before we use them.

MRichards99 (approved):

MRichards99:
@VKTB It looks like https://github.com/Ezard/semantic-prs would a good alternative

VKTB:
All good now, thanks for sorting this out @MRichards99!

```sh
(venv) [datagateway-api]$ semantic-release print-version
8.0.0
```

VKTB (dismissed):

RKrahl:
Dear all, many thanks for working on this!

I tried it out, it clearly shows progress, but we are not there, I'm afraid.

## Test setup

I deployed `datagateway_api` in Apache with `mod_wsgi`. The relevant bits of the config was like:

`httpd.conf`:
```xml
<VirtualHost _default_:443>

        ServerName data.example.org:443

        WSGIDaemonProcess datagateway \
                display-name=%{GROUP} user=wwwrun group=www threads=3
        WSGIScriptAlias /api /usr/lib/python3.6/site-packages/datagateway_api/wsgi.py \
                process-group=datapub application-group=%{GLOBAL}

</VirtualHost>
```
`config.yaml`:
```yaml
datagateway_api:
  extension: "/datagateway"
  # ...
search_api:
  extension: "/panosc"
  # ...
url_prefix: "/api"
```

## Observations

As expected, I could reach the Swagger UI for the datagateway and for search API at `https://data.example.org/api/datagateway` and `https://data.example.org/api/panosc` respectively. It looked good at first glance.

But the endpoints in the swagger UI were listed as:
```
Dataset
GET /panosc/Datasets
GET /panosc/Datasets/{pid}
...
```
and the calls after hitting `Try it out` -> `Execute` were made to `https://data.example.org/panosc/...`. So the Swagger UI was still kind of assuming the application to be deployed at root, dropping the `url_prefix` in the path of all internal URLs.

louise-davies:
@RKrahl sorry about that - I overlooked that the URLs in the swagger UI were wrong. I've fixed it so it now displays the correct URLs - respecting the `url_prefix` - which can then be used to query the API in the swagger UI sucessfully.

RKrahl:
Great! I'll have a look.

RKrahl:
Tested, works, looks good!

PR #416, opened 2023-04-20
Remove outdated paragraph on constraints on the setuptools version

RKrahl:
## Description
Suggest to remove an outdated paragraph from the README: according to `pyproject.toml` you are requiring `python-icat = "1.0.0"`. This version dropped support for Python 2 and does not use `2to3` any more. So there is no (known) constraint on the `setuptools` version in current `python-icat`.

## Testing Instructions
No code changes, only documentation update.


MRichards99:
Good spot! I've removed references to this on our GitHub Actions as we don't need to uninstall & reinstall setuptools, which will speed up the CI times. GitHub Actions has passed so I'll merge this PR

MRichards99 (approved):

PR #426, opened 2023-06-01
Containerize application and configure GitHub Actions to build and push Docker image to Harbor

VKTB:
## Description
This PR adds a Dockerfile which can be used to build a Docker image and create a container using it. It is configured to create a production image and run a Gunicorn server on port `8000` when a container is started.

To add an extra layer of security, the image has been set up so that containers run as non-root.

Environment variables have also been defined in the Dockerfile to allow for values to be passed at runtime to future running containers. These values are then used by the `docker-entrypoint.sh` script, which acts as an entry point script that is executed when the container starts, to update the config values in the `config.yaml` file. The environment variables are:
- `ICAT_URL` (Default value: `http://localhost`)
- `ICAT_CHECK_CERT` (Default value: `false`)
- `LOG_LOCATION` (Default value: `/dev/stdout`)

Finally, this PR adds a `docker` job to the `ci-build` GitHub Actions workflow to build a Docker image using the Dockerfile introduced in this PR and push the same to Harbor. The job only triggers if all the other jobs succeed and does different things depending on the context. It builds the Docker image in all cases and also pushes the image to Harbor only if something is pushed to the `main` branch.

To build an image, run the command below in the `datagateway-api` repo directory
```
docker build -t datagateway_api_image .
```
To start a container on port `8000` from the image that you just built, run:
```
docker run -p 8000:8000 --name datagateway_api_container datagateway_api_image 
```
If you want to pass values for the environment variables then instead run:
```
docker run -p 8000:8000 --name datagateway_api_container --env ICAT_URL=https://127.0.0.1:8181 --env ICAT_CHECK_CERT=true --env LOG_LOCATION=/datagateway-api-run/logs.log datagateway_api_image
```

## Testing Instructions

- [ ] Review code
- [ ] GitHub Actions should build and push an image to Harbor when this PR is merged


MRichards99:
Your CI failures look similar to what we've seen on OG API. Have a look at the PR I made to fix the issues: https://github.com/ral-facilities/operationsgateway-api/pull/59. The Semantic PR job seems to be working so that's good

VKTB:
Thanks @MRichards99. From the PR that you linked, it looks like upgrading Poetry to 1.3.2 fixes this issue. However, I just had a look at Poetry's release notes and noticed that they dropped Python 3.6 support in their 1.2.0 version (see more [here](https://python-poetry.org/history/#changed-11)) and this project still supports 3.6.

.github/workflows/ci-build.yml
```diff
@@ -352,3 +352,35 @@ jobs:
...
+  docker:
+    # This job triggers only if all the other jobs succeed and does different things depending on the context.
+    # The job builds the Docker image in all cases and also pushes the image to Harbor only if something is
+    # pushed to the main branch.
+    needs: [tests, linting, formatting, safety, generator-script-testing]
+    name: Docker
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Check out repo
+      uses: actions/checkout@v2
```
VKTB, line 365: Pin CI actions to commit SHAs (and update them if necessary).

MRichards99:
Ah yes, that's true. Having to support Python 3.6 is becoming a bit of an issue, there's so many dependencies that need upgrading but we can't as future versions drop 3.6 support (as you've experienced with Poetry).

The root of this issue seems to be relating to a dependency called cachecontrol, read https://github.com/psf/cachecontrol/issues/292

You could try pinning `urllib3` as per https://github.com/psf/cachecontrol/issues/292#issuecomment-1535373382 but lots of people saying that doesn't work. You could also try the same with requests as per https://github.com/psf/cachecontrol/issues/292#issuecomment-1535282946 but again, lots saying that doesn't work.

There's a brief discussion where someone is tied to a Poetry version < 1.2 at https://github.com/python-poetry/poetry/issues/7877. I don't fully understand what's being discussed here and how we'd translate that into our repo but could be worth a try if you can work out what to do.

We really need to move to support 3.8+ on this repo, but we can't until we can get 3.8 on a RIG VM/Docker for production 😢

VKTB:
Thank you @MRichards99. I think I somehow fixed the failures in https://github.com/ral-facilities/datagateway-api/pull/427. I agree, though, that we need to drop support for the older Python 3 versions.

ajkyffin (approved):

PR #468, opened 2023-12-21
Fix Pip Install Issues

### LINKED ISSUE #467: Add CI job to test pip install
**Description**:
Currently, `pip install` does not install a working set of dependencies. This should be detected by the CI.

**Acceptance criteria**:
- [ ] A CI job installs datagateway_api into a venv using pip
- [ ] The CI job tests that the installed package runs correctly

MRichards99:
This PR will close #467

## Description
These changes allow DataGateway API to be installed using `pip install` - the fix is simply just to explicitly state which version of Flask should be used instead of the installation process trying to figure that out. Normally that wouldn't be an issue, but we're having to use quite an old version of Flask that still supports Python 3.6. There's also some updates to `poetry.lock` as a result - a lot of that is just upgrading other dependencies as the file hasn't changed in a little while.

I've added a job to the CI to keep track of this - you can see in [this run](https://github.com/ral-facilities/datagateway-api/actions/runs/7289367196) (before Flask was pinned), that new job fails on Python 3.8+, giving the same error that was originally found but the CI now passes.

When the PR is merged, it will create v9.0.2 so you'll be able to use this new version with SciGateway Ansible.

## Testing Instructions
Take a look at the CI and check you're happy with it and try to pip install the API on your own machine.

- [ ] Review code

## Agile Board Tracking
Connect to #467

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -361,11 +361,58 @@ jobs:
       - name: Diff SQL dumps
         run: diff -s ~/generator_script_dump_main.sql ~/generator_script_dump_1.sql
 
+  pip-install-testing:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+    name: Pip Install ${{ matrix.python-version }} Test
+    steps:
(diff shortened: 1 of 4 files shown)


louise-davies (approved):
LGTM - thanks for fixing this! It will be helpful for future machine migrations

PR #473, opened 2024-02-16
Disable search API in docker image

ajkyffin:
## Description
Remove the `search_api` section from `config.yaml` in the docker image, otherwise it fails to load since there is no search API backend available.

## Testing Instructions
- [ ] Review code
- [ ] CI will build a container image for the PR and upload to Harbor. Test in kubernetes.

## Agile Board Tracking
N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,8 +34,14 @@ RUN --mount=type=cache,target=/root/.cache \
     ln -s "$DATAGATEWAY_API_LOCATION/datagateway_api/" datagateway_api; \
     \
     # Create config.yaml and search_api_mapping.json from their .example files \
-    cp datagateway_api/config.yaml.example datagateway_api/config.yaml; \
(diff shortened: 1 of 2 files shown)


louise-davies:
Why do we need to disable the search api? We'd need it to deploy in production for ISIS at least, and probably eventually for DLS when they have open data.

ajkyffin:
> Why do we need to disable the search api?
Unless there's a search backend available, datagateway-api won't start unless it's disabled.

ajkyffin:
It seems I've been confused between the search api and the new search functionality being worked on in ICat. When the container was failing to load with errors about the search API, I assumed it was because the new search functionality wasn't present. In fact the problem was I had `/icat` in the icat_url which isn't needed for python-icat.

Anyway, once it's pointing to an ICat server correctly it starts fine. Whether or not it works without the PANOSC scoring API I don't know.

PR #478, opened 2024-03-22
Pin python docker images to SHA

VKTB:
## Description
This PR pins the Python docker images in the Dockerfiles to commit SHAs so that Dependabot can create PRs each time there is a change in the upstream base images that the Python image is set to use as new SHAs are generated on every upstream change. Unfortunately, Dependabot only updates the Python versions but not the OS versions and there is an open issue for this. This means that we will have to do the Alpine version updates manually for now. This PR also configures Dependabot to ignore Python major and minor versions so again we will have to do such Python upgrades manually when we are ready to do that.

## Testing Instructions
- [ ] Review code

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,9 @@ updates:
       interval: daily
       time: "08:00"
       timezone: "Europe/London"
+    ignore:
+      - dependency-name: "python"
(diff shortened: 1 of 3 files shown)


joshdimanteto:
This pr is now outdated. The implementation of pinning the python version in the docker images and updating the dependabot config has been done in these prs #569 and #586

PR #488, opened 2024-09-20
Implement Reader Performance Queries for /datafiles and /datasets

MRichards99:

## Description
These changes allow efficiency gains to be made when either the datafiles or datasets are being queried, with a WHERE filter for an ID of the parent entity (e.g. WHERE dataset.id = 4 on /datafiles). The changes have been implemented for the count endpoints of datafiles and datasets.

If a relevant request comes in, the API queries the parent entity directly, to check whether the user can see it (i.e. has the permissions to). If the user can, their original query is sent to ICAT (JPQL query constructed by Python ICAT as normal) but it is executed as the 'reader' user (as configured in `config.yaml`). This should bypass the additional complexity added to queries by ICAT Server, which is causing performance issues. As I don't have rules setup in my ICAT, I wasn't able to test this, but hopefully testing next week done by others should test that.

This functionality is completely optional - if the config isn't in `config.yaml` or it is disabled, the API will behave as normal (i.e. pass user queries onto ICAT as the user of the API request).

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Connect to #{issue number}


kevinphippsstfc (changes requested):
Having reviewed the code, made a number of changes to it and tested it on preprod, I'm happy that it is functioning as required. I think there are just a few things to complete the code to a "production" level: complete the TODOs, add docstrings to each function, and add some tests for the new functionality. That last one may not be trivial given that it probably requires starting up a new instance of the API using a different config.

MRichards99:
This is ready for review. I've addressed the TODOs, added documentation in the form of docstrings and added a test class to test the new functionality. Quite a bit of time went into fixing the CI which was broken, presumably because a PR hasn't gone through this repo for a little while. I've also edited a couple of the commit messages so a new release will be made. 

The generator script CI job fails but this is intentional - final diff in the CI will still fail because `main` doesn't contain two dropped columns that I've added to an SQL script, but the initial diff to check that two runs produce identical data passes. This is what I expected to happen :)

kevinphippsstfc (approved):
Looks good! 

Thanks for tidying up the code and adding the tests.

And thanks for implementing this bit of functionality which appears to have been successful in reducing database load.

PR #521, opened 2026-01-07
Upgrade to pydantic v2 #520

### LINKED ISSUE #520: Upgrade to pydantic V2
- update datagateway api config models
- update search api config models

joshdimanteto:
This PR will close #520

## Description
- Update pydantic to version 2
- Change field_validator that are use repetitively to use annotated pattern

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Connect to #520


datagateway_api/src/common/config.py
```diff
@@ -173,22 +166,22 @@ def load(cls, path=None):
         except (IOError, ValidationError) as error:
             sys.exit(f"An error occurred while trying to load the config data: {error}")
 
-    @validator("search_api")
-    def validate_api_extensions(cls, value, values):  # noqa: B902, N805
+    @field_validator("search_api")
+    def validate_api_extensions(cls, value, info):  # noqa: B902, N805
```
patrick-austin, line 168:
```suggestion
    @field_validator("search_api")
    @classmethod
    def validate_api_extensions(cls, value, info):  # noqa: B902, N805
```

Based on the docs, I think field validators are supposed to be class methods? There's a mix of `cls` and `self` below but I think it should be the former.

datagateway_api/src/common/config.py
```diff
@@ -173,22 +166,22 @@ def load(cls, path=None):
...
         """
         Checks that the DataGateway API and Search API extensions are not the same. An
         error is raised, at which point the application exits, if the extensions are the
         same.
 
-        :param cls: :class:`APIConfig` pointer
+        :param self: :class:`APIConfig` pointer
```
patrick-austin:
```suggestion
        :param cls: :class:`APIConfig` pointer
```

datagateway_api/src/search_api/models.py
```diff
@@ -1,20 +1,69 @@
...
+    str,
+    BeforeValidator(normalize_id),
+    Field(alias="id"),
+]
+
+SearchAPIIdOptional = Annotated[
+    Optional[str],
+    BeforeValidator(normalize_id),
+    Field(None, alias="id"),
+]
```
patrick-austin: I'm not sure what the point of the validator is here - is it an artifact of a previous attempt at this or is it actually doing something functional? If the `type` is `str` (rather than `StrictStr`, where I can see this doing something), then Pydantic would coerce `int`, `float` or `bool` anyway on init wouldn't it?

In terms of `SearchAPIIdOptional` I'd also find `Optional[SearchAPIId]` (in keeping with local style) or `SearchAPIId | None` (my personal preference as it find it more explicit) more readable than aliasing a second type, but that's subjective.

Haven't had a chance to actually play with this so I might be missing something - will try to run this when I'm back in the office).

datagateway_api/src/search_api/models.py
```diff
@@ -223,16 +268,12 @@ class Dataset(PaNOSCAttribute):
...
-    @root_validator(pre=True)
-    def set_is_public(cls, values):  # noqa: B902, N805
+    @model_validator(mode="before")
+    def set_is_public(cls, value):  # noqa: B902, N805
         # Hardcoding this to True because anon user is used for querying so all data
         # returned by it is public
-        values["isPublic"] = True
-        return values
+        value["isPublic"] = True
+        return value
```
patrick-austin: Would it be simpler to just set `is_public: Literal[True]` on line 260 rather than doing this?

datagateway_api/src/search_api/models.py
```diff
@@ -263,16 +304,12 @@ class Document(PaNOSCAttribute):
...
```
patrick-austin: Again can we achieve this by using literal types rather than root validators?

patrick-austin (commented):
Had a few thoughts on whether some of the pydantic functions are actually required or whether they can be simplified, but I've not actually run the code so some of my suggestions might be wrong (it's based on vague memory of how validation etc. works).

Will try and do some experimentation to check these suggestions make sense at some point, but it might be quicker for you to have a go if you have time?

datagateway_api/src/search_api/models.py
joshdimanteto: The validator is needed because of how from_icat() builds the models. That method pulls values straight from the ICAT response into entity_data without converting ID fields, so ICAT’s integer IDs are still ints when we call cls(**entity_data), and Pydantic rejects them for fields typed as str.

The BeforeValidator is the only point where those integer IDs get normalised into strings before model construction, so removing it causes from_icat() to fail.

Optional[SearchAPIId] isn’t equivalent to the aliased optional type, because SearchAPIId alone doesn’t include that normalisation step.

datagateway_api/src/search_api/models.py
joshdimanteto: Using Literal[True] alone isn’t enough here because the issue isn’t validating the value, it’s that from_icat() may not include isPublic in entity_data at all when building the PaNOSC model. Literal[True] only checks the value after it's present, but it doesn’t supply or inject the field if the ICAT mapping didn’t produce it at that point.


The model_validator(before) ensures isPublic is explicitly added to the incoming data before model construction, which avoids the “Field required” validation error. So we still need the validator to guarantee the field exists during the from_icat() path; Literal[True] doesn’t replace that behaviour.

patrick-austin:
```python
>>> class Model(BaseModel):
...     b: Literal[True] = True
... 
>>> Model().model_dump()
{'b': True}
>>> Model(b=True).model_dump()
{'b': True}
>>> Model(b=False).model_dump()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/kez92691/icat-doi-minter-api/.venv/lib64/python3.11/site-packages/pydantic/main.py", line 250, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for Model
b
  Input should be True [type=literal_error, input_value=False, input_type=bool]
    For further information visit https://errors.pydantic.dev/2.12/v/literal_error
```

You can still set the default value - unless this still wouldn't give what you need for some reason? I suppose the current behaviour wouldn't fail validation if you pass in a value of `False`, but the above would?

datagateway_api/src/search_api/models.py
patrick-austin: > To deal with this issue, the type of such PaNOSC fields were changed to str instead as this casts integers to strings but still throws a ValidationError if None is passed to it.

I was sort of taking this on faith, but looking at your models outside of the lines changed, you're not setting [coerce_numbers_to_str](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.coerce_numbers_to_str)? The statement above will only be true if you actually set this, otherwise the strict mode distinction won't change anything. For example (minimal, but hopeful this is still representative):
```python
from pydantic import BaseModel, ConfigDict, Field

class Model(BaseModel):
    id_: str = Field(alias="id")
    
    @classmethod
    def from_icat(cls):
        return cls(**{"id": 1})
    
    model_config = ConfigDict(coerce_numbers_to_str=True)

print(Model.from_icat().model_dump())
```

joshdimanteto: I thought this would convert all numbers in the model to strings but looking at this example (https://stackoverflow.com/a/77618503) it doesn't look like it. it is only if type is a string . I've made the change and it works. Cheers

datagateway_api/src/search_api/models.py
joshdimanteto: Your suggestion is the clean solution, From my understand if it was false this code would have to be updated anyways as it assumed to be always true as it querying using the anonymous user

patrick-austin (approved):
Glad those suggestions did actually work / make sense in the end. I haven't done any targeted manual testing (just checked that the API starts and I can submit an arbitrary query OK), since this didn't add/change any specific functionality and the tests are passing on the CI. I'm happy to approve based on that but let me know if you think there's something worth manually checking in more detail.

Issue #519, opened 2025-11-05
Port API to FastAPI


joshdimanteto:
**Description**:
Flask-RESTful is no longer maintained so the API should be ported to FastAPI.

PR #523, opened 2026-01-09
Refactor dg api model to pydantic #522

### LINKED ISSUE #522: refactor datagateway api model to use pydantic V2
The datagateway api models are currently generated using sqlalchemy. These need to be changed to use pydantic V2 before the repo is refactored to use fast api #519    

- all sql imports should be removed

joshdimanteto:
This PR will close #522

## Description
Refactor the SQL models to use pydantic 
Generate the models use the SQL received from the icat server 
Fix the models for the post and patch requests 

The fields in the models are all optional, as get endpoint uses the distinct filter which only returns the fields selected. Also this simplifies the cascade logic as  all validation will happen on ICAT. 

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Closes #522


datagateway_api/src/datagateway_api/build_models.py
```diff
@@ -0,0 +1,131 @@
...
+    mod_id: str = Field(alias="modId")
+    mod_time: datetime = Field(alias="modTime")
+
+
+def build_datagateway_api_model(**kwargs):
+    """
+    Build the datagateway models using the SQL scheme given by the ICAT server
+
+    :returns dict of name and pydantic model key value pair
+    """
```
joshdimanteto, line 93:
```suggestion

"""
    Dynamically construct Pydantic models for all ICAT entities exposed by the
    connected ICAT server.

    This function queries the ICAT server for its schema (entity names, fields,
    types, relationships, and nullability) and generates a set of Pydantic
    models representing:

    - The base entity model for each ICAT entity (e.g. `Investigation`)
    - A corresponding POST model for creation (e.g. `InvestigationPost`)
    - A corresponding PATCH model for partial updates (e.g. `InvestigationPatch`)

    Relationship fields (ONE or MANY) are converted into model references or
    lists of ICAT IDs. Attribute fields are mapped to Python/Pydantic primitive
    types according to the `TYPE_MAP`. Optionality and nullability are respected
    for all generated fields. Field descriptions from ICAT (if present) are
    carried forward into the model metadata.

    All generated models are finally rebuilt (`model_rebuild`) using the full
    model namespace so that forward references between models resolve correctly.

    Parameters
    ----------
    **kwargs :
        Optional configuration parameters. Expected keys:
        - `client_pool`: A pool or cache of ICAT clients, passed into
          `get_cached_client`.

    Returns
    -------
    dict
        A dictionary mapping model names (e.g. `"Investigation"`,
        `"InvestigationPost"`, `"InvestigationPatch"`) to their corresponding
        dynamically generated Pydantic model classes.

    Raises
    ------
    PythonICATError
        If the ICAT server reports an error while fetching entity names or
        entity schema information.

    Notes
    -----
    - Models include metadata (via `Annotated[... , Field(...)]`) for descriptions.
    - SYSTEM_FIELDS are always excluded from the generated models.
    - Relationship fields use forward references and are resolved at the end of
      generation.
    - The POST and PATCH models differ by optionality and update semantics.

    """
```

datagateway_api/src/datagateway_api/icat/helpers.py
```diff
@@ -573,6 +573,23 @@ def create_entities(client, entity_type, data):
...
+                        # Even when forcing it to work by using the wrong type, the
+                        # GET request still does not return the one-to-one related
+                        # values (e.g. "Facility f INCLUDE f.parameterTypes").
+                        #
+                        # After attempting to fix the GET behaviour, the create
+                        # operation now throws a duplicate reference error when
+                        # saving related entities.
+                        #
+                        # Fix list handling, one-to-one include behaviour, and
+                        # duplicate reference errors.
```
joshdimanteto, line 592: This is fix in #519

datagateway_api/src/datagateway_api/build_models.py
```diff
@@ -0,0 +1,165 @@
...
+    except ICATError as e:
+        raise PythonICATError(e) from e
+
+    for name in entity_names:
+        info = client.getEntityInfo(name)
+        fields = {}
+        post_fields = {}
+        for field in info.fields:
+            post_name = f"{name}Post"
+            patch_name = f"{name}Patch"
```
patrick-austin: If this doesn't depend on `field`, can it be refactored into the outer loop over `entity_names`?

datagateway_api/src/datagateway_api/build_models.py
```diff
@@ -0,0 +1,165 @@
...
+                description = getattr(field, "comment", None)
+                field_metadata = Field(description=description)
+                optional_annotated_type = Annotated[optional_field_type, field_metadata]
+
+                fields[field.name] = (optional_annotated_type, None)
+                post_fields[field.name] = (optional_annotated_type, None)
+
+            else:
+                rel_model_name = field.type
+                post_type = None
```
patrick-austin:
```suggestion
```
Both branches of the `if` block lead to this being assigned, so don't need to assign to None.

datagateway_api/src/datagateway_api/build_models.py
```diff
@@ -0,0 +1,165 @@
...
+                if field.relType == "MANY":
+                    rel_type_str = f"List['{rel_model_name}']"  # noqa: B907
+                    post_type = f"List['{rel_model_name}Post']"  # noqa: B907
+                else:
+                    rel_type_str = f"'{rel_model_name}'"  # noqa: B907
+                    post_type = Optional[int]
+
+                optional_type = Optional[post_type]
```
patrick-austin:
```suggestion
                    post_type = int

                optional_type = Optional[post_type]
```

`Optional[Optional[int]]` still evaluates to `Optional[int]`, and we only ever use `post_type` here, so I think it can be simplified this way.

patrick-austin (changes requested):
I haven't had a chance to try running this yet (will try to tomorrow), but there's a couple of minor inefficiencies / lines that I think can be removed.

patrick-austin (approved):

PR #528, opened 2026-02-23
feat(query): add isnull operator for NULL checks

### LINKED ISSUE #527: No where filter operator supports null checking / Add is null & is not null operators
Related to the DLS DOI work. The difference between an open or closed DataPublication in that project is whether `publicationDate` is defined or not. So, in order to be able to have an open/closed filter, we need to be able to query whether `publicationDate` is null or not null. However, there's no operator for this, as using the `eq`/`=` operator doesn't work, you need to use `IS` (i.e. `IS NULL`).

I don't really have an idea for what to name the operator, maybe `isnull` and `nisnull` to keep in with the rest of the operators (where n is the negative version) or maybe just `isnull` and `notisnull`. Or maybe just add it as the `IS` operator, so `is` and `nis` - this would help if we need to query on boolean columns in the future.

joshdimanteto:
This PR will close #527 

## Description
see #527 

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
#527


datagateway_api/src/swagger/initialise_spec.py
```diff
@@ -157,6 +162,8 @@ def initialise_datagateway_api_spec(spec):
             "examples": {
                 "eq": {"value": [{"id": {"eq": 1}}]},
                 "ne": {"value": [{"id": {"ne": 1}}]},
+                "is": {"value": [{"publicationDate": {"is": "NULL"}}]},
+                "nis": {"value": [{"complete": {"nis": True}}]},
```
joshdimanteto:
```suggestion
                "isnull": {"value": [{"publicationDate": {"isnull": True}}]},
```

patrick-austin (commented):
Code looks OK to me but haven't had a chance to run it yet, will try and do that when I can before approving.

I note that this is not the solution that @louise-davies suggested - are you happy enough with this as is and dealing with a general solution for the discrepancy between boolean equality and is-ness (not sure what the word for that is - identity?) when/if it comes to it?

louise-davies:
> Code looks OK to me but haven't had a chance to run it yet, will try and do that when I can before approving.
> 
> I note that this is not the solution that @louise-davies suggested - are you happy enough with this as is and dealing with a general solution for the discrepancy between boolean equality and is-ness (not sure what the word for that is - identity?) when/if it comes to it?

@patrick-austin Apparently JPQL doesn't support the `IS` operator, just specific `IS NULL` (and also `IS EMPTY`) so it had to be the specific `isnull` operator. What I was looking up was SQL specific 😓

patrick-austin (approved):
Tried it on my dev ICAT and gave the expected number of results for `"isnull": true` and `"isnull": false`.

PR #533, opened 2026-03-11
feat: Add ability to supply settings file via volume #532

### LINKED ISSUE #532: Add ability to supply settings file via volume
https://github.com/ral-facilities/operationsgateway/blob/develop/docker/docker-entrypoint.sh
https://github.com/ral-facilities/datagateway/blob/develop/docker/docker-entrypoint.sh

joshdimanteto:
This PR will close #532

## Description
 Add ability to supply settings file via volume similar to data gateway and operations gateway 

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
closes #532

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Dockerfile to build and serve datagateway-api
 
 # Build stage
-FROM python:3.11-alpine3.17 AS builder
+FROM python:3.11-alpine@sha256:f07e2ace46f560f09a6eeec7b4913b80ee99546e749ef82342a419a326620856 AS builder
 
 WORKDIR /datagateway-api-build
 
@@ -16,7 +16,7 @@ RUN --mount=type=cache,target=/root/.cache \
 
 
 # Install & run stage
-FROM python:3.11-alpine3.17
+FROM python:3.11-alpine@sha256:f07e2ace46f560f09a6eeec7b4913b80ee99546e749ef82342a419a326620856
 
 WORKDIR /datagateway-api-run
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Dockerfile to build and serve datagateway-api
 
 # Build stage
-FROM python:3.11-alpine3.17 as builder
+FROM python:3.11-alpine3.17 AS builder
 
 WORKDIR /datagateway-api-build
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -41,8 +41,8 @@ RUN --mount=type=cache,target=/root/.cache \
     addgroup -S datagateway-api; \
     adduser -S -D -G datagateway-api -H -h /datagateway-api-run datagateway-api; \
     \
-    # Change ownership of config.yaml - the entrypoint script will need to edit it \
-    chown datagateway-api:datagateway-api datagateway_api/config.yaml;
+    # Change ownership of settings location - the entrypoint script will need to edit it \
+    chown -R datagateway-api:datagateway-api datagateway_api/;
```
joshdimanteto, line 45: Changed to match what was done in operationsgateway and datagateway

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Dockerfile to build and serve datagateway-api
 
 # Build stage
-FROM python:3.11-alpine3.17 as builder
+FROM python:3.11-alpine3.17 AS builder
```
louise-davies: Whilst we're here modifying the dockerfile - should do a couple things:
1. Should pin the images to specific SHA
2. Should either upgrade the `alpine` tag to latest `alpine3.23` or just use `python:3.11-alpine` to ensure we're tracking the latest alpine

Would also need to do this for the run stage as well

docker/docker-entrypoint.sh
```diff
@@ -1,17 +1,21 @@
...
-rm "$TEMPFILE"
+    # Set values in config.yaml from environment variables
+    # No quotes for icat_check_cert because it's boolean
+    sed -e "s|icat_url: \".*\"|icat_url: \"$ICAT_URL\"|" \
+        -e "s|icat_check_cert: .*|icat_check_cert: $ICAT_CHECK_CERT|" \
+        -e "s|log_location: \".*\"|log_location: \"$LOG_LOCATION\"|" \
+        /datagateway-api-run/datagateway_api/config.yaml > "$TEMPFILE"
+
+    cat "$TEMPFILE" > /datagateway-api-run/datagateway_api/config.yaml
+    rm "$TEMPFILE"
```
louise-davies: Missing end to if statement
```suggestion
fi
```

Also - I'm getting errors from docker because I think this file has windows line endings - I needed to convert it to linux ones to get it to work - so can that be fixed? Either I can commit the file I have working or you can fix it @joshdimanteto

louise-davies (changes requested):

docker/docker-entrypoint.sh
joshdimanteto: I've addressed the changes in the latest commit. I don't get error when i build the image locally so if it is still an issue in the latest commit for you, can you push your changes @louise-davies

louise-davies: It's probably an issue on my end with git line endings etc. as the repo for me is on windows so I probably just need to fiddle with my git config

louise-davies (approved):
LGTM - locally was able to build the image and supply the settings via a volume 🎉

Issue #517, opened 2025-02-26
Remove older python version


joshdimanteto:
Remove older versions of python  "3.6", "3.7", "3.8"
#515

Issue #526, opened 2026-02-12
Use pydantic settings to manage config


joshdimanteto:
https://github.com/ral-facilities/datagateway-api/pull/507#discussion_r2787719122

joshdimanteto:
https://github.com/ral-facilities/datagateway-api/pull/534#discussion_r2924101859

PR #534, opened 2026-03-11
Release/v10.0.0

joshdimanteto:
## Description

Merging develop into main for V10.0.0 release.

- #532
- #517
- #527

## Testing Instructions

- [ ] Review code
- [ ] run`poetry run semantic-release print-version` locally to ensure next version is 10.0.0


.github/workflows/ci-build.yml
```diff
@@ -152,73 +134,70 @@ jobs:
...
   linting:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     name: Linting
     steps:
       - name: Setup Python
         uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
         with:
-          python-version: "3.9.7"
+          python-version: "3.10.4"
```
patrick-austin: `3.10.4` is 4 years old - is there a specific reason to pin to this, given we use `3.10` elsewhere?

.github/workflows/ci-build.yml
```diff
@@ -152,73 +134,70 @@ jobs:
...
   formatting:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     name: Code Formatting
     steps:
       - name: Setup Python
         uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
         with:
-          python-version: "3.9.7"
+          python-version: "3.10.4"
```
patrick-austin: See above.

.github/workflows/ci-build.yml
```diff
@@ -152,73 +134,70 @@ jobs:
...
   safety:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     name: Dependency Safety
     steps:
       - name: Setup Python
         uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
         with:
-          python-version: "3.9.7"
+          python-version: "3.10.4"
```
patrick-austin: See above.

.github/workflows/ci-build.yml
```diff
@@ -229,12 +208,12 @@ jobs:
...
         uses: actions/setup-java@b36c23c0d998641eff861008f374ee103c25ac73 # v4.4.0
         with:
-          distribution: 'temurin'
+          distribution: "temurin"
           java-version: 8
       - name: Setup Python
         uses: actions/setup-python@f677139bbe7f9c59b41e40162b753c062f5d49a3 # v5.2.0
         with:
-          python-version: "3.9.7"
+          python-version: "3.10.4"
```
patrick-austin: See above.

datagateway_api/src/common/filter_order_handler.py
```diff
@@ -87,7 +88,7 @@ def add_icat_relations_for_non_related_fields_of_panosc_related_entities(
         python_icat_include_filter = None
         icat_relations = []
         for filter_ in self.filters:
-            if type(filter_) == PythonICATIncludeFilter:
+            if type(filter_) is PythonICATIncludeFilter:
```
patrick-austin, line 91: Generally speaking [it's better to use ](https://www.geeksforgeeks.org/python/type-isinstance-python/)`isinstance(a, b)` than `type(a) is b` - is there a reason for this (i.e. NOT wanting to return `true` for subclasses)?

Dockerfile
patrick-austin, line 1: The Dockerfile uses Python3.11, but the README.md and CI use 3.10? Is there a reason to not make this consistent (especially given Python3.10 is "only" LTS until October).

datagateway_api/src/datagateway_api/database/backend.py
```diff
@@ -61,7 +61,7 @@ def login(self, credentials, **kwargs):
             )
             return session_id
         else:
-            raise AuthenticationError("Username and password are incorrect")
+            raise AuthenticationError("Username and password are incorrect") from None
```
patrick-austin: I don't think `from None` serves a purpose here as we are not in the context of an `except` block? What is it intended to achieve?

datagateway_api/src/search_api/panosc_mappings.py
```diff
@@ -11,9 +11,14 @@
 
 class PaNOSCMappings:
     def __init__(
-        self, path=Path(__file__).parent.parent.parent / "search_api_mapping.json",
+        self,
+        path=None,
```
patrick-austin: This could now fit on one line.

docker/docker-entrypoint.sh
patrick-austin, line 1: I'm not sure how this is intended to work (but I'm not great at shell). If `/datagateway-api-run/datagateway_api/config.yaml` doesn't exist, then the sed command will fail too? 

Regardless, I don't think this should be necessary. By default Pydantic settings will take env variables as a possible configuration source and can use them to overwrite lower priority sources, in particular default settings - so if `$ICAT_URL` etc. are defined and pydantic is configured correctly you _should_ get this for free?

test/integration/datagateway_api/test_backends.py
```diff
@@ -17,7 +17,7 @@ class TestBackends:
     def test_valid_backend_creation(self, backend_name, backend_type):
         test_backend = create_backend(backend_name)
 
-        assert type(test_backend) == backend_type
+        assert type(test_backend) is backend_type
```
patrick-austin: Again, is there a reason to not use `isinstance`?

test/unit/search_api/test_panosc_mappings.py
```diff
@@ -152,13 +162,16 @@ def test_valid_get_panosc_non_related_field_names(
         test_panosc_entity_name,
         expected_non_related_field_names,
     ):
-        non_related_field_names = test_panosc_mappings.get_panosc_non_related_field_names(  # noqa: B950
-            test_panosc_entity_name,
+        non_related_field_names = (
+            test_panosc_mappings.get_panosc_non_related_field_names(  # noqa: B950
```
patrick-austin:
```suggestion
            test_panosc_mappings.get_panosc_non_related_field_names(
```

By my count this line is now only 69 characters, so does it need to ignore B950?

test/unit/test_get_entity_object.py
```diff
@@ -14,18 +14,22 @@ class TestGetEntityObject:
...
+                type(FACILITY),
+                id="plural entity name, 'y' to 'ies'",
             ),
         ],
     )
     def test_valid_get_entity_object_from_name(self, entity_name, expected_object_type):
         database_entity = get_entity_object_from_name(entity_name)
 
-        assert type(database_entity) == expected_object_type
+        assert type(database_entity) is expected_object_type
```
patrick-austin: `isinstance`?

util/icat_db_generator.py
```diff
@@ -876,16 +903,20 @@ def generate_dataset(self, i):
...
         dataset.investigation = self.client.get(
             "Investigation",
-            investigation_id
-            if investigation_id != 0
-            else InvestigationGenerator.amount - 1,
+            (
+                investigation_id
+                if investigation_id != 0
+                else InvestigationGenerator.amount - 1
+            ),
```
patrick-austin, line 910: I find the ternary operators for the modular id division (particularly when spread over 5 lines) not very easy to read. I was going to suggest a change here, but frankly, the entire generation script is undermined by the decision to use `range(1, N)` since this excludes `N` from the range. So you have these off by one errors all over the place, and all the comments saying there are so many of this entity are going to be wrong.

Changing all that would be out scope, so on balance I'm OK with this as it is.

test/integration/search_api/test_search_api_query_filter_factory.py
```diff
@@ -1781,15 +1849,15 @@ def test_valid_include_filter_with_where_filter_in_scope(
...
             if isinstance(test_filter, NestedWhereFilters):
                 for expected_nested in expected_nested_wheres:
                     assert repr(test_filter) == repr(expected_nested)
-                    expected_nested_wheres.remove(expected_nested)
+                    expected_nested_wheres.remove(expected_nested)  # noqa: B909
             if isinstance(test_filter, SearchAPIWhereFilter):
                 for expected_where in expected_where_filter_data:
                     assert repr(test_filter) == repr(expected_where)
-                    expected_where_filter_data.remove(expected_where)
+                    expected_where_filter_data.remove(expected_where)  # noqa: B909
```
patrick-austin: I don't think you should be ignoring the linter here, removing elements from the list inside the for loop will effectively mean you skip every second assert. Which is giving you what you want because (and only because) none of your expected lists have more than 2 entries, and what you actually want to do is `break` the loop when a match is found. But the moment one of these lists have 3 entries, this is going to fail as it will assert on entries 1 and 3, and the second assert will fail. You're also vulnerable to the case where the right number of filters are returned of the wrong type, since it may enter the first block, but because `expected_included_entities` is an empty list we don't assert anything at all.

There is no need for the inner `for` loops here at all. What you should be doing is something like:
```python
expected_where = expected_where_filter_data.pop(0)
assert repr(test_filter) == repr(expected_where)
```
Provided the order of filters is determinate (which it must be in order for this code to pass) I think this should work.

I might have missed something here just going off the code and not having run it, but as is, this looks brittle to me (that's what the linter is telling you and should not be ignored). At least it uses `isinstance` though 👍🏼

test/integration/search_api/test_search_api_query_filter_factory.py
```diff
@@ -1867,15 +1935,15 @@ def test_valid_include_filter_with_include_filter_in_scope(
...
+            test_request_filter,
+            test_entity_name,
         )
 
         assert len(filters) == expected_length
         for test_filter in filters:
             if isinstance(test_filter, SearchAPIIncludeFilter):
                 for expected_include in expected_included_entities:
                     assert test_filter.included_filters == expected_include
-                    expected_included_entities.remove(expected_include)
```
patrick-austin, line 1878: As above, removing the `remove` is only a solution because all the lists happen to be single entry. I think it would be better (i.e. more resilient to future test changes as well as more explicit/understandable) to `pop(0)` from `expected_included_entities` in the outer loop and get rid of the inner.

test/integration/search_api/test_search_api_query_filter_factory.py
```diff
@@ -2059,31 +2135,27 @@ def test_valid_filter_input_with_all_filters(
...
                     assert repr(test_filter) == repr(expected_where)
-                    expected_where_filter_data.remove(expected_where)
             if isinstance(test_filter, SearchAPILimitFilter):
                 for expected_limit in expected_limit_values:
                     assert test_filter.limit_value == expected_limit
-                    expected_limit_values.remove(expected_limit)
             if isinstance(test_filter, SearchAPISkipFilter):
                 for expected_skip in expected_skip_values:
                     assert test_filter.skip_value == expected_skip
-                    expected_skip_values.remove(expected_skip)
```
patrick-austin, line 2156: See above.

datagateway_api/src/datagateway_api/database/models.py
```diff
@@ -44,21 +43,23 @@ def __init__(self, enum_type):
     def process_bind_param(self, value, dialect):
         if isinstance(value, self.enum_type):
             return value.value
-        raise DatabaseError(f"value {value} not in {self.enum_type.__name__}")
+        raise DatabaseError(f"value {value} not in {self.enum_type.__name__}") from None
```
patrick-austin: See other comment - we don't have any except context to suppress?

patrick-austin (changes requested):
Most of this is fine (whitespace changes)/I agree with (chaining exceptions), but there are a few points that I think should be changed:
* Inconsistency in Python version
* `isinstance`
* `from None` being used outside of `except`?
* `docker/docker-entrypoint.sh‎`
* Removing entities from within the loop

docker/docker-entrypoint.sh
louise-davies: The intent of the shell script is to check if the config file exists, if it does then it doesn't use the example config and edit the environment variables into it. aka this means if a config file has been supplied via a volume then the script should leave it alone, if one isn't provided then it does what it previously did which was the templating.

I don't know about pydantic, because this uses a YAML config I presumed I couldn't configure it via env variables and needed to provide the `config.yaml` hence the change. If pydantic provides the ability to do defaults via env variables then it would be good to see an example compose file that I could adapt.

patrick-austin: Yeah that intent makes sense but
https://github.com/ral-facilities/datagateway-api/blob/52297aeee6b71c6aad358c4f342cc0eb0bb89a65/docker/docker-entrypoint.sh#L4
is `true` if config doesn't exist
and
https://github.com/ral-facilities/datagateway-api/blob/52297aeee6b71c6aad358c4f342cc0eb0bb89a65/docker/docker-entrypoint.sh#L12-L15
takes the config file as input - but that doesn't exist so won't you just get `No such file or directory` in the case where the config is not provided?

> I don't know about pydantic, because this uses a YAML config I presumed I couldn't configure it via env variables and needed to provide the config.yaml hence the change. If pydantic provides the ability to do defaults via env variables then it would be good to see an example compose file that I could adapt.

Right now it won't work because (I just checked now) this repo doesn't use the pydantic settings class (?) but just manually loads a pydantic model. If you did, you can [prioritise your sources](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#customise-settings-sources):
Loading from config.env (low priority):
```
>>> from icat_doi_minter_api.config import Settings
>>> Settings().facility
'LILS'
```
Loading from actual environment, if defined (mid priority):
```
$ export FACILITY="LILS2"
```
```
>>> from icat_doi_minter_api.config import Settings
>>> Settings().facility
'LILS2'
```
Loading from init (highest priority):
```
>>> from icat_doi_minter_api.config import Settings
>>> Settings(facility="LILS3").facility
'LILS3'
```

louise-davies: Ah, I didn't notice that about that it expects the `config.yaml` to exist. It should instead use the `config.example.yaml` (it previously copied the example file earlier, hence why it edits `config.yaml` here)

It sounds like improving the settings can come in a later PR/branch? As what's here works for now (other than changing the sed command to act on `config.example.yaml`)

.github/workflows/ci-build.yml
joshdimanteto: I'll update this to use 3.10 like the other places

docker/docker-entrypoint.sh
joshdimanteto: I'll create an task for this

util/icat_db_generator.py
joshdimanteto: I'll create an issue of this

test/integration/search_api/test_search_api_query_filter_factory.py
joshdimanteto: I tried to refactor this test when I was doing this #517 pr but the test keep on failing. I'll try your solution to see if this fixes it 👍

joshdimanteto: It work 🥳, I'm implementing the change now

docker/docker-entrypoint.sh
joshdimanteto: #526

datagateway_api/src/common/filter_order_handler.py
joshdimanteto: There’s a comment just below explaining why we don’t in this case. here is the comment 

```
                # Using `type` as `isinstance` would return `True` for any class that
                # inherits `PythonICATIncludeFilter` e.g. `SearchAPIIncludeFilter`.`
```

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Dockerfile to build and serve datagateway-api
 
 # Build stage
-FROM python:3.11-alpine@sha256:f07e2ace46f560f09a6eeec7b4913b80ee99546e749ef82342a419a326620856 AS builder
+FROM python:3.10.4-alpine@sha256:sha256:3ccf8d308cd8cf6b97540a2c5b202b8ce513bdd43a1fd73c1d9e657b96e7ded4 AS builder
```
patrick-austin:
```suggestion
FROM python:3.10.4-alpine@sha256:3ccf8d308cd8cf6b97540a2c5b202b8ce513bdd43a1fd73c1d9e657b96e7ded4 AS builder
```

Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN --mount=type=cache,target=/root/.cache \
 
 
 # Install & run stage
-FROM python:3.11-alpine@sha256:f07e2ace46f560f09a6eeec7b4913b80ee99546e749ef82342a419a326620856
+FROM python:3.10.4-alpine@sha256:sha256:3ccf8d308cd8cf6b97540a2c5b202b8ce513bdd43a1fd73c1d9e657b96e7ded4
```
patrick-austin:
```suggestion
FROM python:3.10.4-alpine@sha256:3ccf8d308cd8cf6b97540a2c5b202b8ce513bdd43a1fd73c1d9e657b96e7ded4
```

patrick-austin (changes requested):

Dockerfile
joshdimanteto: Cheers the version was also wrong as well. I need to use this version as it does not have an vulnerabilities  

https://hub.docker.com/layers/library/python/3.10.14-alpine/images/sha256-155d65335d7dddeaaeb109ff7deefb36405ed0829064577c38446356ba1e9c14

`python:3.10.14-alpine@sha256:3ccf8d308cd8cf6b97540a2c5b202b8ce513bdd43a1fd73c1d9e657b96e7ded4`

joshdimanteto: ignore what i said above i should be using this https://hub.docker.com/layers/library/python/3.10.20-alpine/images/sha256-d13464d245790978bfe05464d5ed04907929032e5488b1b6351e869c8ad27572

patrick-austin (approved):

docker/docker-entrypoint.sh
```diff
@@ -1,17 +1,21 @@
...
+    TEMPFILE="$(mktemp)"
 
-cat "$TEMPFILE" > datagateway_api/config.yaml
-rm "$TEMPFILE"
+    # Set values in config.yaml from environment variables
+    # No quotes for icat_check_cert because it's boolean
+    sed -e "s|icat_url: \".*\"|icat_url: \"$ICAT_URL\"|" \
+        -e "s|icat_check_cert: .*|icat_check_cert: $ICAT_CHECK_CERT|" \
+        -e "s|log_location: \".*\"|log_location: \"$LOG_LOCATION\"|" \
+        /datagateway-api-run/datagateway_api/config.yaml > "$TEMPFILE"
```
patrick-austin: This still needs to change to `.example` or equivalent right?

patrick-austin (changes requested):

docker/docker-entrypoint.sh
louise-davies:
```suggestion
        /datagateway-api-run/datagateway_api/config.yaml.example > "$TEMPFILE"
```

patrick-austin (approved):

PR #558, opened 2026-05-22
Update the release process #557

### LINKED ISSUE #557: Update the release process
- Remove the automatic releases 
- Use same process to IMS https://github.com/ral-facilities/datagateway-api/wiki/Release-Process

joshdimanteto:
## Description
see #557 , #555

- Remove safety, the plan is to use dependabot for vulnerabilities  
- Change release process to manual instead of automatic. This is following what has been use for IMS.

1.  tags should be pushed to harbor when a release has been made 
2.  docker image for branches can be optionally pushed to harbor from the actions tag 
3. https://github.com/ral-facilities/datagateway-api/wiki/Release-Process 

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
closes #557 ,#555

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1,22 +1,43 @@
 name: Release Build
 on:
   push:
-    branches:
-      - main
+    tags: "v*"
+
+permissions:
+  contents: read
 
 jobs:
-  build:
-    name: Release Build
+  docker:
+    # This job builds the Docker image and if successful, it pushes it to Harbor.
+    name: Docker
     runs-on: ubuntu-latest
-
     steps:
-      - name: Checkout repo
-        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
-        with:
-          fetch-depth: 0
-          token: ${{ secrets.ADMIN_PAT }}
+      - name: Check out repo
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - name: Python Semantic Release
-        uses: python-semantic-release/python-semantic-release@c1bcfdbb994243ac7cf419365d5894d6bfb2950e # v9.12.0
+      - name: Login to Harbor
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
         with:
-          github_token: ${{ secrets.ADMIN_PAT }}
+          registry: ${{ vars.HARBOR_URL }}
+          username: ${{ secrets.HARBOR_USERNAME }}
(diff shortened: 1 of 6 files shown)


.github/workflows/release-build.yml
```diff
@@ -1,22 +1,43 @@
...
+            latest=false
 
-      - name: Python Semantic Release
-        uses: python-semantic-release/python-semantic-release@c1bcfdbb994243ac7cf419365d5894d6bfb2950e # v9.12.0
+      - name: Build and push Docker image to Harbor
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
         with:
-          github_token: ${{ secrets.ADMIN_PAT }}
+          context: .
+          file: ./Dockerfile.prod
```
joelvdavies:
```suggestion
```
There isn't a Dockerfile.prod, and it doesnt have a prod stage, so I think should be removed to use the Dockerfile as is.

joelvdavies (commented):
Haven't tested the poetry/the nox change, just looked at the CI workflows comparing to IMS, noticed one issue.

.github/workflows/release-build.yml
```diff
@@ -1,22 +1,43 @@
...
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # v5.10.0
         with:
-          fetch-depth: 0
-          token: ${{ secrets.ADMIN_PAT }}
+          images: ${{ vars.HARBOR_URL }}/datagateway-api
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
```
joelvdavies, line 32: Also just realised, currently there are released versions under both `X.Y.Z` & `vX.Y.Z` tags. As this matches IMS, its currently configured to use `X.Y` & `X.Y.Z`, so in our deployment config we use X.Y and can avoid modifying it for patch updates. Not sure if this is intentional or not, but whoever is pulling them should be made aware of the change if this is kept.

joshdimanteto: @louise-davies

louise-davies: I think I tend to use the specific `vX.Y.Z` tags but that's just because that's what was there/matches the github tag name. However I can change the configs to use the `X.Y.Z` tags, and we can delete the old `v` tags on harbor to reduce confusion.

Equally the only place this is deployed is on the dev machine, so not much needs changing!

patrick-austin (approved):
Changes look OK to me, input for the workflow appears in the GitHub UI: 

Image

PR #561, opened 2026-06-02
Fix/swagger include query param #560

joshdimanteto:
## Description
Swagger UI was incorrectly serialising the `include` query parameter when
defined with a complex `oneOf` schema (string, array, object). This caused:

- object values to be flattened into query params
- arrays to be split into multiple parameters
- nested structures to break entirely

Although the backend supports these formats correctly, the Swagger UI
could not represent or send them properly.

Fix:
- Force `include` to be treated as a string in OpenAPI using:
  - schema.type=string
- Convert examples to JSON strings so they are transmitted correctly
- Fix string examples 

Result:
- Swagger now sends `include` as a single encoded JSON value
- All supported formats work consistently
- Backend logic unchanged

## Testing Instructions
For the GET investigations endpoint:

example 1: 
order : "id asc"
skip : 0
limit: 50
include:  {"investigationInstruments":"instrument"}

example 2: 
order : "id asc"
skip : 0
limit: 50
include: "investigationInstruments"

example 3: 
order : "id asc"
skip : 0
limit: 50
include: ["investigationInstruments", "publications"]

example 3: 
order : "id asc"
skip : 0
limit: 50
include: ["publications", {"investigationInstruments":"instrument"}]

example 5: 
order : "id asc"
skip : 0
limit: 50
include:  {"investigationInstruments":["instrument","investigation"]}

example 6: 
order : "id asc"
skip : 0
limit: 50
include:  {"investigationInstruments":{"instrument":"facility"}}

- [x] Review code
- [ ] {more steps here}

## Agile Board Tracking
closes #560

### DIFF
--- a/datagateway_api/src/datagateway_api/routers/entity.py
+++ b/datagateway_api/src/datagateway_api/routers/entity.py
@@ -109,7 +109,7 @@ OrderQuery = Query(
     title="ORDER_FILTER",
     description=("Apply order filters to the query. Given a field and direction, order the returned entities."),
     json_schema_extra={"type": "array", "items": {"type": "string", "default": ""}},
-    openapi_examples={"asc": {"value": ["id asc"]}, "desc": {"value": ["id desc"]}},
+    openapi_examples={"asc": {"value": ['"id asc"']}, "desc": {"value": ['"id desc"']}},
 )
 LimitQuery = Query(
     default=None,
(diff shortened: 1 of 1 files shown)


patrick-austin (approved):
Tried it with various includes of increasing complexity (`["investigationInstruments.instrument.instrumentScientists", {"datasets": ["type", "datafiles.datafileFormat"]}]`) and it all seemed to work as I'd expect.

PR #563, opened 2026-06-09
chore: remove nox

joshdimanteto:
## Description
This PR removes Nox from the repository. Because the project is now pinned to a single Python version, a multi-version test runner is no longer necessary
Changes:

- Removed noxfile.py.
- Updated CI workflows to run tests directly via pytest , flake8 and black.
- Updated ReadME

## Testing Instructions

- [ ] Review code

## Agile Board Tracking
closes #566


.github/workflows/ci-build.yml
patrick-austin, line 1: During the `tests` job, we run `poetry install` 3 times (lines 54, 119, 127). Is this really necessary, or can we just call this once before the first time we need it?

patrick-austin (commented):
Functionally it looks OK to me, but I think we might be calling `poetry install` repeatedly when we only need it once in the CI? Unless there's something too subtle for me to pick up on going on here.

.github/workflows/ci-build.yml
joshdimanteto: Yeah this can get called once, I've updated this in the latest commit.

patrick-austin (approved):

PR #569, opened 2026-06-17
migrate to use uv #568

joshdimanteto:
## Description

migrate to use UV
- update docker file 
- update readme

Followed these steps:
github.com/ral-facilities/inventory-management-system-api/pull/758#issuecomment-4345268687

## Testing Instructions

- [ ] Review code

## Agile Board Tracking
closes #568


joshdimanteto:
icatdb Generator Script Consistency Test

This test won't pass as uv is not on develop

.github/workflows/ci-build.yml
```diff
@@ -31,17 +31,9 @@ jobs:
...
-      # We will not be facing this issue when we upgrade to a newer Poetry version.
-      - name: Install Requests
-        run: pip install 'requests<2.30'
-
-      # Install dependencies
-      - name: Install dependencies
-        run: poetry install
+      # Install uv and API's dependencies
+      - name: Install uv
+        run: pip install uv
```
patrick-austin: Given we were pinning to a specific version of `poetry`, should we be pinning `uv`?

.pre-commit-config.yaml
```diff
@@ -1,17 +1,17 @@
...
+      - id: check-yaml
+      - id: detect-private-key
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: local
     hooks:
-    -   id: black
+      - id: black
         name: black
         entry: poetry run black
```
patrick-austin:
```suggestion
        entry: uv run black
```

patrick-austin (changes requested):

.github/workflows/ci-build.yml
joshdimanteto: https://github.com/astral-sh/setup-uv#usage

image


image

.github/workflows/ci-build.yml
joelvdavies, line 29: I would add a `.python-version` file with 3.11 in it, then it can be used for dev too. Then use the setup-uv action instead to let uv handle it.
If you want to pin the uv vesion you can use
```
with:
    version: "0.11.21"
```

patrick-austin (approved):
Happy with the changes I requested, but don't know if you're planning on doing the python version file Joel requested (couldn't see it in changes since my last review).

.github/workflows/ci-build.yml
joshdimanteto: The Python setup is required to create the ICAT server from the icat-ansible GitHub repository. This will change when I dockerise the development, testing, and linting environments, as I plan to set up the ICAT server using a Docker container on harbor

joshdimanteto:
> Happy with the changes I requested, but don't know if you're planning on doing the python version file Joel requested (couldn't see it in changes since my last review).

It should be done in the latest commit, I didn’t realise the .python-version file was in .gitignore.

joelvdavies (approved):

VKTB (approved):
Sorry, I don't have the time to test this myself so only looked at the changes and they LGTM. From my migration to uv PRs, I found that the prod Docker stage was successfully building but errored on container start up due to stuff I missed so I would suggest that you test running a container with the production image to ensure it's working.

VKTB:
Clicked the approve button too quick but I can see that some dependencies in the `pyproject.toml` file are pined to a minimum (`>=`) but some to exact (`==`) versions, what's the reason behind this?

Also, what is the `apispec` dependency for?

joshdimanteto:
> Clicked the approve button too quick but I can see that some dependencies in the `pyproject.toml` file are pined to a minimum (`>=`) but some to exact (`==`) versions, what's the reason behind this?
> 
> Also, what is the `apispec` dependency for?

The apispec will be removed as this is not needed anymore since the FastAPI upgrade, good spot. No they should all have minimums I'll change this

joshdimanteto:
> Sorry, I don't have the time to test this myself so only looked at the changes and they LGTM. From my migration to uv PRs, I found that the prod Docker stage was successfully building but errored on container start up due to stuff I missed so I would suggest that you test running a container with the production image to ensure it's working.

Cheers, I’ve tested this using the config file mounted as a volume and it works as expected.

The environment variable approach does not currently work on the develop branch, even when following the instructions in the README. due to #533

I’m planning to remove the environment variable handling from the production stage, as it adds unnecessary complexity and is no longer needed.

Going forward, the expectation is that users will provide configuration by mounting the required files as volumes. This includes config.yaml, logging.ini, and search_api_mapping.json.

All configuration should be managed directly through these files rather than being modified at runtime by an entrypoint script.

VKTB:
@joshdimanteto When I was doing this a while back, I wasn't quite sure why we need to have the image create "default" files from the example ones. I understand that it is useful for local development but for production they will always need to be modified/overridden because the example files do not have the required configuration values for production environment. I prefer the container to fail to start because it's missing the mandatory files instead of it creating default ones - removing these like you have done and expecting for the files to be mounted via volumes is the better thing to do in my opinion.

VKTB (approved):

Issue #540, opened 2026-03-16
Simplify query parameter handling


joshdimanteto:
https://github.com/ral-facilities/datagateway-api/pull/524#discussion_r2939457267

PR #573, opened 2026-06-22
Migrate to ruff

joshdimanteto:
## Description
Migrate to use ruff

 `PLW2901` was ignored within the datagateway-api directory as the test coverage is 43%. This should handle separately to prevent tests failing.  #574

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
closes #570


datagateway_api/common/filters.py
```diff
@@ -20,7 +20,7 @@ def apply_filter(self, query):
 class WhereFilter(QueryFilter):
     precedence = 1
 
-    def __init__(self, field, value, operation):
+    def __init__(self, _field, value, operation):
```
patrick-austin, line 23: I'm not sure how I feel about this. I assume since the first argument is unused, it wants it to be marked with an underscore? But now the arguments for `WhereFilter` are different from its subclasses `PythonICATWhereFilter` and `SearchAPIWhereFilter`. And we never directly instantiate `WhereFilter` as far as I can see (one place we `import PythonICATWhereFilter as WhereFilter` just to be extra confusing. And `SearchAPIWhereFilter` is a subclass of `PythonICATWhereFilter` not `WhereFilter`. So why does `WhereFilter` even exist?

Similarly, the other contents of `ral-facilities/datagateway-api/datagateway_api/search_api/filters.py` all import from `ral-facilities/datagateway-api/datagateway_api/datagateway_api/icat/filters.py` not `ral-facilities/datagateway-api/datagateway_api/common/filters.py`. Can we just remove this entire file and migrate the shared functionality to `ral-facilities/datagateway-api/datagateway_api/datagateway_api/icat/filters.py` classes? Appreciate that's scope creep so should be in another PR. Maybe I'm missing something subtle too.

In the case of this line I'd leave it and just ignore the QA rather than changing it if it's going to get nuked later, but either way it's not a big deal. The important bit to me is: why does this file exist?

datagateway_api/common/helpers.py
```diff
@@ -102,11 +101,11 @@ def get_filters_from_query_string(request: Request, api_type, entity_name=None):
     :return: The list of filters
     """
     if api_type == "search_api":
-        from datagateway_api.search_api.query_filter_factory import (
+        from datagateway_api.search_api.query_filter_factory import (  # noqa: PLC0415
             SearchAPIQueryFilterFactory as QueryFilterFactory,
         )
     elif api_type == "datagateway_api":
-        from datagateway_api.datagateway_api.query_filter_factory import (
+        from datagateway_api.datagateway_api.query_filter_factory import (  # noqa: PLC0415
```
patrick-austin, line 108: I don't like ignoring these, as I don't like doing conditional imports mid-function. I also don't like the fact that a `common` helper has to import from two of our api specific modules at all.

Given we pass `api_type` to distinguish between the two cases, would  it be acceptable (and to my mind simpler) to just pass either `SearchAPIQueryFilterFactory` or `DataGatewayAPIQueryFilterFactory` as an argument of type `class`, and call it directly below?

datagateway_api/datagateway_api/icat/helpers.py
```diff
@@ -415,21 +411,18 @@ def execute_entity_query(client, entity_type, filters, aggregate=None):
...
+    config file and 'enabled' in that section is set to true.
     """
     reader_config = Config.config.datagateway_api.use_reader_for_performance
     if not reader_config:
         return False
-    if not reader_config.enabled:
-        return False
-
-    return True
+    return reader_config.enabled
```
patrick-austin, line 419: I'd be tempted to say this should be `return reader_config is not None and reader_config.enabled`, but also plan  on  changing this to get rid of `enabled` altogether as soon as I can get back to working on the reader stuff so fine as is.

datagateway_api/datagateway_api/routers/entity.py
```diff
@@ -206,12 +205,12 @@ def get_endpoint(
...
-        limit: int = LimitQuery,  # pylint:disable=unused-argument
-        skip: int = SkipQuery,  # pylint:disable=unused-argument
-        distinct: List[str] = DistinctQuery,  # pylint:disable=unused-argument
-        include: Any = IncludeQuery,  # pylint:disable=unused-argument
+        where: list[Json] = WhereQuery,  # noqa: ARG001
+        order: list[str] = OrderQuery,  # noqa: ARG001
+        limit: int = LimitQuery,  # noqa: ARG001
+        skip: int = SkipQuery,  # noqa: ARG001
+        distinct: list[str] = DistinctQuery,  # noqa: ARG001
+        include: Any = IncludeQuery,  # noqa: ARG001
```
patrick-austin, line 213: See #540 - implementing that would (I hope) resolve the fact we don't directly use any of these variables.

datagateway_api/datagateway_api/routers/entity.py
```diff
@@ -409,9 +408,9 @@ def get_count_endpoint(
     )
     def get(
         request: Request,
-        where: List[Json] = WhereQuery,  # pylint:disable=unused-argument
-        distinct: List[str] = DistinctQuery,  # pylint:disable=unused-argument
-        include: Any = IncludeQuery,  # pylint:disable=unused-argument
+        where: list[Json] = WhereQuery,  # noqa: ARG001
+        distinct: list[str] = DistinctQuery,  # noqa: ARG001
+        include: Any = IncludeQuery,  # noqa: ARG001
```
patrick-austin, line 413: See #540

datagateway_api/datagateway_api/routers/entity.py
```diff
@@ -459,12 +458,12 @@ def get_find_one_endpoint(
...
-        limit: int = LimitQuery,  # pylint:disable=unused-argument
-        skip: int = SkipQuery,  # pylint:disable=unused-argument
-        distinct: List[str] = DistinctQuery,  # pylint:disable=unused-argument
-        include: Any = IncludeQuery,  # pylint:disable=unused-argument
+        where: list[Json] = WhereQuery,  # noqa: ARG001
+        order: list[str] = OrderQuery,  # noqa: ARG001
+        limit: int = LimitQuery,  # noqa: ARG001
+        skip: int = SkipQuery,  # noqa: ARG001
+        distinct: list[str] = DistinctQuery,  # noqa: ARG001
+        include: Any = IncludeQuery,  # noqa: ARG001
```
patrick-austin, line 466: See #540

datagateway_api/datagateway_api/build_models.py
```diff
@@ -129,13 +128,12 @@ def build_datagateway_api_model(**kwargs):
...
         patch_name = f"{name}Patch"
         for field in info.fields:
-
             if field.name in SYSTEM_FIELDS:
                 continue
 
             if field.relType == "ATTRIBUTE":
                 field_type = TYPE_MAP.get(field.type, str)
-                optional_field_type = Optional[field_type]
+                optional_field_type = Optional[field_type]  # noqa: UP045
```
patrick-austin, line 136: Does this not work if you do `field_type | None`?

datagateway_api/datagateway_api/build_models.py
```diff
@@ -153,7 +151,7 @@ def build_datagateway_api_model(**kwargs):
                     rel_type_str = f"{rel_model_name!r}"
                     post_type = int
 
-                optional_type = Optional[post_type]
+                optional_type = Optional[post_type]  # noqa: UP045
```
patrick-austin, line 154: Does this not work if you do `post_type | None`?

datagateway_api/datagateway_api/query_filter_factory.py
```diff
@@ -16,9 +26,9 @@
 
 class DataGatewayAPIQueryFilterFactory(QueryFilterFactory):
     @staticmethod
-    def get_query_filter(request_filter, entity_name=None):
+    def get_query_filter(request_filter, _entity_name=None):
```
patrick-austin, line 29: I'm not sure I feel comfortable with the abstract and DG version of this having `_entity_name` while Search has `entity_name`. If the motivation is to have a consistent method signature (see docstring below) then we now don't have this, and have a redundant argument in two places. Furthermore the search version also has `related_entity_name`, so they didn't even have consistent signatures before.

As far as I can see there's only one place (that helper function with the mid-function imports) that even uses something that might be either DG or search. Everywhere else calls one of the classes directly? I'm not sure what the best approach here is. Possibly that these should all take `**kwargs` and then `entity_name` and `related_entity_name` can be safely passed and ignored, and explicitly defined for the search implementation which actually uses them? Assuming that `ruff` is satisfied, which it might not be.

Bottom line is I agree with `ruff` that we shouldn't have `entity_name` as an argument, but I'm not convinced the best solution is just to mark it with `_` in some places but not others.

datagateway_api/search_api/routers/entity.py
```diff
@@ -145,7 +145,7 @@ def get_search_endpoint(
     @search_api_error_handling
     def get(
         request: Request,
-        filter_: str = FilterQuery,  # pylint:disable=unused-argument
+        filter_: str = FilterQuery,  # noqa: ARG001
```
patrick-austin, line 148: I'm going to stop tagging #540 on these - suffice to say I hope we can remove all `noqa: ARG001` when that's actioned.

datagateway_api/search_api/helpers.py
```diff
@@ -59,7 +58,7 @@ def assign_status_code(e, status_code):
         try:
             # If no status code exists (for non-API defined exceptions), assign a status
             # code
-            e.status_code
+            e.status_code  # noqa: B018
```
patrick-austin, line 61: I don't think we should be ignoring the qa here if we can help it. Using `try/except` for things that can be `if` statements is probably bad practice. If possible, can we replace the block with:
```python
if not hasattr(e, status_code):
    e.status_code = status_code
```
Maybe that won't work, in which case we go with the ignore. But I'd hope we can make this more explicit and avoid the qa altogether.

datagateway_api/search_api/nested_where_filters.py
```diff
@@ -91,10 +89,10 @@ def set_search_api_query(query_filter, search_api_query):
     def __str__(self):
         """
         Join the condition on the left with the one on the right with the boolean
-        operator
+        operator.
         """
         boolean_algebra_list = [self.lhs, self.rhs]
-        try:
+        try:  # noqa: SIM105
```
patrick-austin, line 95: Could we just not add `[None]` to the list in the first place and avoid the whole `try/except`? E.g.:
```python
boolean_algebra_list = [i for i in [self.lhs, self.rhs] if i != [None]]
```
We already have list comprehension one liners below...

test/integration/datagateway_api/icat/endpoints/test_count_with_filters_icat.py
```diff
@@ -38,7 +38,8 @@ def test_valid_no_results_count_with_filters(
         valid_icat_credentials_header,
     ):
         test_response = test_client.get(
-            '/datagateway-api/investigations/count?where={"title": {"like": "This filter should cause 0 results to be found '
+            '/datagateway-api/investigations/count?where={"title": {"like": "This filter should cause 0 results to be '
+            "found "
             'for testing purposes..."}}',
```
patrick-austin:
```suggestion
            'found for testing purposes..."}}',
```

test/integration/datagateway_api/icat/endpoints/test_get_with_filters_icat.py
```diff
@@ -26,7 +26,9 @@ def test_valid_no_results_get_with_filters(
         valid_icat_credentials_header,
     ):
         test_response = test_client.get(
-            '/datagateway-api/investigations?where={"title": {"eq": "This filter should cause a 404 for testing purposes..."}}',
+            '/datagateway-api/investigations?where={"title": {"eq": "This filter should cause a 404 for testing '
+            "purposes"
+            '..."}}',
```
patrick-austin:
```suggestion
            'purposes..."}}',
```
Will it not let us have this? If not then fine, but 3 lines is too many IMHO.

util/icat_db_generator.py
```diff
@@ -688,10 +690,10 @@ class DataPublicationGenerator(Generator):
     amount = InvestigationGenerator.amount
 
     def generate(self):
-        for i in range(1, self.amount):
-            DataPublicationGenerator().generate_data_publication(i)
+        for _ in range(1, self.amount):
+            DataPublicationGenerator().generate_data_publication(self)
```
patrick-austin:
```suggestion
            DataPublicationGenerator.generate_data_publication(self)
```
Think this is a typo? We only need to pass `self` if calling statically (which we do in the above examples).
Alternatively:
```python
            DataPublicationGenerator().generate_data_publication()
```
Whichever feels most consistent across the file.

util/icat_db_generator.py
```diff
@@ -1103,7 +1105,7 @@ def generate_datafile_parameter(cls, i):
         datafile_param.create()
 
 
-def generate_all(i, generators, client):
+def generate_all(i, generators, _client):
```
patrick-austin: Why not just remove this argument entirely given we don't use it?

patrick-austin (changes requested):
Broadly speaking I think the changes to the linting has exposed a few places where in my opinion improvements could be made. In a few of these I'm not convinced that simply ignoring the warning or even following `ruff`'s suggestion is the right approach, as it is correctly flagging things which are problematic and should (if possible) be changed more significantly.

I completely appreciate that:
1. some of these may be too far out of scope for this PR
2. it's a matter of opinion, and something that fails my vibe check may pass someone else's

In that respect happy for some of these to be separate issues/nested PRs/discussed further or with others before action.

joshdimanteto:
> Broadly speaking I think the changes to the linting has exposed a few places where in my opinion improvements could be made. In a few of these I'm not convinced that simply ignoring the warning or even following `ruff`'s suggestion is the right approach, as it is correctly flagging things which are problematic and should (if possible) be changed more significantly.
> 
> I completely appreciate that:
> 
>     1. some of these may be too far out of scope for this PR
> 
>     2. it's a matter of opinion, and something that fails my vibe check may pass someone else's
> 
> 
> In that respect happy for some of these to be separate issues/nested PRs/discussed further or with others before action.

Most of the changes you suggested i will create separate issues for. With the current coverage at 43% I don't think it is worth the risk of changing things within this pr without a more in depth analysis. This issue #577 is going to the parent issue of all the inner refactor that need to be addressed within dg-api. The aim of this is to do this whilst also added more tests to ensure it works as expected.

datagateway_api/common/filters.py
joshdimanteto: This might be a remnant of the database backend potentially

joshdimanteto: I'll create an issue for this

util/icat_db_generator.py
joshdimanteto: yes, thanks i didn't even realise i added in the parenthesis

test/integration/datagateway_api/icat/endpoints/test_get_with_filters_icat.py
joshdimanteto: Good spot, this must have been done when i allowed it to do automatic fixes

datagateway_api/common/helpers.py
joshdimanteto: I'll create an issue for this

datagateway_api/datagateway_api/build_models.py
joshdimanteto: No it gives this error when you try to run dg-api. 

```
    optional_type = post_type | None
                    ~~~~~~~~~~^~~~~~
TypeError: unsupported operand type(s) for |: 'str' and 'NoneType'
```

patrick-austin: OK looking at the full context it's because `post_type` might be the class `int` (line 152, fine) or it might be the string literal `f"List['{rel_model_name}Post']"` (line 150, not fine). I don't know if we can get away with `list[f"{rel_model_name}Post"` instead on line 150, as it's not clear to me exactly what we're using this type for. There are limitations to subscripted generics (i.e. we can't do `isinstance(None, List["ModelPost"] | None)`, but equally we wouldn't be able to do that with `isinstance(None, Optional['List["ModelPost"]'])` either, so 🤷🏼

datagateway_api/datagateway_api/query_filter_factory.py
joshdimanteto: I am going to change this to ignore this warning and then I'll create an issue to fix this here and in all the other places

PR #576, opened 2026-06-23
chore(deps): group dependent bot PRs

joshdimanteto:
## Description
Configure Dependabot to manage Docker, GitHub Actions, and uv dependencies across the develop and main branches. Updates are scheduled regularly with a 7-day cooldown. GitHub Actions and uv dependencies are grouped into single PRs to simplify review and maintenance.

Quoted from IMS repo

_We use Renovates default of 3 days for npm packages on the frontend, here it seems for pip 7 days is generally considered preferable, so went for that. Doesnt resolve supply chain attacks, but relies on the community to notice something is wrong which typically occurs within a short timespace after packages are released._

## Testing Instructions

- [ ] Review code

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,76 @@
 version: 2
 updates:
+  # Maintain Dockerfile dependencies on the `develop` branch
+  - package-ecosystem: "docker"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


joshdimanteto:
#586

PR #607, opened 2026-07-14
feat(docker): dockerise dev workflow (test, lint, format, ICAT, app)

joshdimanteto:
## Description
Fix count on search-api tests due to improper clean up when creating a new investigation 
This pr dockerise the development process for datagateway-api. This has been done to be similar to https://github.com/ral-facilities/icat-doi-minter-api/blob/main/docker-compose.yaml .  

## Testing Instructions

- [ ] Review code

## Agile Board Tracking
Closes #612

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -17,375 +17,177 @@ jobs:
     strategy:
       fail-fast: false
     name: Build & Tests
+
     steps:
-      - name: Add apt repo
-        run: sudo add-apt-repository universe
-
-      # Setup Python
-      - name: Setup Python
-        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
-        with:
-          python-version: "3.11"
-          architecture: x64
-
-      - name: Install uv
-        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
-        with:
-          version: "0.11.21"
-
       - name: Checkout DataGateway API
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
-      # Prep for using the API for tests
-      - name: Configure datagateway extension
-        run: echo "`yq \
-          '.datagateway_api.extension="/datagateway-api"' datagateway_api/config.yaml.example`" > datagateway_api/config.yaml.example
-      - name: Create config.yaml
-        run: cp datagateway_api/config.yaml.example datagateway_api/config.yaml
-      - name: Create search_api_mapping.json
-        run: cp datagateway_api/search_api_mapping.json.example datagateway_api/search_api_mapping.json
-      - name: Create logging.ini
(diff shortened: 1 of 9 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,202 @@
...
+        condition: service_healthy
+    ports:
+      - "14747:4848" # payara port
+      - "18181:8181" # https port
+      - "18080:8080" # https port
+    volumes:
+      - ./icat/post-boot-commands.asadmin:/config/post-boot-commands.asadmin
+    environment:
+      - POSTBOOT_COMMANDS=/config/post-boot-commands.asadmin
+      - ROOT_USERS="simple/root anon/anon"
```
patrick-austin: Lines 31/32 imply that adding `anon/anon` to the root users was done at build time, then that image tagged with `dg-api`. Is this environment variable still used?

docker-compose.yml
```diff
@@ -0,0 +1,202 @@
...
+
+    # volumes:
+    ## use for local development
+    # - ./datagateway_api:/datagateway-api-run/datagateway_api
+    depends_on:
+      icat_payara:
+        condition: service_healthy
+    networks:
+      - dg_network
+    profiles: [tests]
```
patrick-austin: Despite the name, this doesn't seem to be calling `pip` anywhere? So I'm thinking:
- Do we still want a test for this? (probably?)
- If so, is it more appropriate to use the `test` target or the `prod` target? (my feeling is the latter,  since I can't think it needs anything dev-y for this and the closer it is to prod the better the test is?)
- Should the title for this be changed to not reference `pip`? (probably?)

.github/workflows/ci-build.yml
```diff
@@ -17,375 +17,177 @@ jobs:
...
-        run: cp datagateway_api/search_api_mapping.json.example datagateway_api/search_api_mapping.json
-      - name: Create logging.ini
-        run: cp datagateway_api/logging.example.ini datagateway_api/logging.ini
-
-      # Launch API to see if it starts correctly or has a startup issue
-      # Code logic used from https://stackoverflow.com/a/63643845
-      - name: Start API
-        run: timeout 10 python -m datagateway_api.main || code=$?; if [[ $code -ne 124 && $code -ne 0 ]]; then exit $code; fi
+      - name: Run pip install test
+        run: docker compose run pip-install-test
```
patrick-austin: See other comment about the name `pip-install-testing` and apparent lack of `pip`?

patrick-austin (commented):
Changes look OK to me, but I had a couple of lines where I'm not sure whether we need it as is or it's an artifact from the previous version that could/should be changed?

docker-compose.yml
joshdimanteto: No, this needs to be removed. I didn't notice this. Thanks

docker-compose.yml
joshdimanteto: I might revert this as this test was need due to this issue which was reported here https://github.com/ral-facilities/datagateway-api/issues/467

patrick-austin (approved):

PR #628, opened 2026-08-07
Implement read-only-api as a new sub API

patrick-austin:

## Description

## Testing Instructions

- [ ] Review code
- [ ] {more steps here}

## Agile Board Tracking
Connect to #{issue number}


datagateway_api/common/config.py
```diff
@@ -60,24 +60,43 @@ class UseReaderForPerformance(BaseModel):
...
+
+    @model_validator(mode="after")
+    def _validate(self) -> Self:
+        if self.default > self.maximum:
+            raise ValueError("default limit cannot exceed maximum limit")
+
+        return self
+
+
+class ReadOnlyAPI(DataGatewayAPI):
```
joshdimanteto: Good use of inheritance here. It may be worth having the Search API model inherit from the DataGateway model too, so they all follow this pattern, which would improve consistency.

joshdimanteto: It might also be worth renaming "DataGatewayAPIExtension" to something like "BaseAPIExtension". The current name could be confusing, as it suggests the type is specific to the datagateway-api sub-API, whereas it appears to be used more generally. A more generic name could make its purpose clearer.

joshdimanteto:
Should this be a multi select ? 
image

joshdimanteto:
Shouldn't this error, as dataset 61 - investigation id should be 1 instead of 2 



https://github.com/user-attachments/assets/b49f6133-3337-4001-b718-2cedfcfc3a7b

patrick-austin:
> Should this be a multi select ? 

I think it should be, as there's nothing wrong with including more than one related entity. Or are you saying that you think it should be, but it wasn't working as a multi-select? Not running the API right now (will check when I can) but I remember being able to select more than one (maybe shift or ctrl clicking)?

> Shouldn't this error, as dataset 61 - investigation id should be 1 instead of 2

So because I'm re-using the existing `get_with_id` function, we don't actually use the parent id as part of the query (just the final id):
https://github.com/ral-facilities/datagateway-api/blob/820c5056dd64cc6886061e4c166d71f1825ce37c/datagateway_api/read_only_api/routers/entities.py#L128-L140
It's preferable to do this than to use the reader account because the (small) overhead of the reader permissions checks takes longer than a primary key (id) query in my tests. Like I said earlier, the hierarchical layout of the endpoints is mostly just for show, but appreciate that it is weird for it to not be used at all in these queries. I suppose could either:
* Re-do the endpoints so they don't have these redundant ids (i.e. have no more than one id in a given path so `/investigations/{id}/datasets/` stays but `/investigations/{id}/datasets/{id}` becomes `/datasets/{id}`
* Add some post query validation that the provided id matches

Happy to go with what you prefer, but if it were up to me I'd probably go for the former (since it's less code).

joshdimanteto:
> > Should this be a multi select ?
> 
> I think it should be, as there's nothing wrong with including more than one related entity. Or are you saying that you think it should be, but it wasn't working as a multi-select? Not running the API right now (will check when I can) but I remember being able to select more than one (maybe shift or ctrl clicking)?

This worked thanks. using ctrl clicking

joshdimanteto (changes requested):
Can you update the `config.yaml.example` please

joshdimanteto:
Could also add these two lines in the docker-compose please ? I forgot to add this in my pr 
image

patrick-austin:
> Could also add these two lines in the docker-compose please ?

Yeah I'd actually also done that myself locally as well but didn't want to commit it since it wasn't in your PR so assumed it was a workaround because I hadn't built the image as intended - lol

joshdimanteto (approved):

